PR #288, opened 2026-07-22
docs: point quickstart at GHCR 1.7.0 and make the version badge dynamic

HimadriSur:
## Why

The README quickstart pulls `public.ecr.aws/f9f6l5u4/samyama-graph:1.1.0`, but our **CI publishes to GHCR, not ECR**. Two registries with entirely separate update paths, which is why the documented quickstart has drifted six releases behind.

Verified tags as of today:

| Registry | Tags |
|---|---|
| `ghcr.io/samyama-ai/samyama-graph` | `latest`, `1.7.0`, `1.7`, `1` |
| `public.ecr.aws/f9f6l5u4/samyama-graph` | `1.1.0` only |

The v1.7.0 **Publish Docker Image** run succeeded, so `1.7.0` is now on GHCR as a proper multi-arch index (`linux/amd64` + `linux/arm64`), and `latest` resolves to the same digest. Pointing the quickstart at GHCR means it now tracks the registry releases actually go to, instead of going stale on every release.

This also matters right now because our [Dev.to launch post](https://dev.to/samyama-ai/your-graphrag-stack-is-two-databases-it-should-be-one-31c6) tells readers to run the GHCR image — as written, anyone following the post to this repo hits contradictory instructions on arrival.

## What changed

- `docker pull` and the Compose `image:` → `ghcr.io/samyama-ai/samyama-graph:1.7.0`
- Prerequisite line no longer says "No AWS account" (no longer an AWS image)
- Version badge `version-1.1.0` (hardcoded) → `img.shields.io/github/v/release/...`, so it reads the latest release and cannot drift again

4 lines. No other content touched.

## Note for maintainers

If ECR is meant to stay the canonical distribution channel, the fix is the other direction — publish versioned images to ECR from CI — and I'm happy to close this and open that instead. Either way the two should stop disagreeing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
     <strong>The graph database that queried 1 billion edges for $2.50</strong>
   </p>
   <p align="center">
-    <a href="https://github.com/samyama-ai/samyama-graph/releases"><img src="https://img.shields.io/badge/version-1.1.0-blue" alt="Version"></a>
+    <a href="https://github.com/samyama-ai/samyama-graph/releases"><img src="https://img.shields.io/github/v/release/samyama-ai/samyama-graph?label=version&color=blue" alt="Version"></a>
     <a href="https://github.com/samyama-ai/samyama-graph/actions"><img src="https://img.shields.io/badge/tests-2238_passing-brightgreen" alt="Tests"></a>
     <a href="LICENSE"><img src="https://img.shields.io/badge/license-Apache_2.0-blue" alt="License"></a>
     <a href="https://graph.samyama.cloud/book/"><img src="https://img.shields.io/badge/book-read_the_docs-orange" alt="Book"></a>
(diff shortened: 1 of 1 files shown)


HimadriSur:
Closing this — the team has confirmed that the **public AWS ECR image is the intended distribution channel** going forward, and that GHCR was the earlier approach. That answers the open question in the PR description, so the change is not wanted. Thanks for the quick call.

Recording the current registry state here so it isn't lost, since this PR was the only place it was written down:

| Registry | Tags today |
|---|---|
| `public.ecr.aws/f9f6l5u4/samyama-graph` | `1.1.0` |
| `ghcr.io/samyama-ai/samyama-graph` | `latest`, `1.7.0`, `1.7`, `1` |

Two things follow from ECR being canonical, both for the team rather than for this PR:

1. **ECR has no image newer than `1.1.0`**, so the README quickstart currently hands new users a build from six releases back. Publishing versioned images to ECR as part of the release process would close that permanently.
2. **CI currently publishes to GHCR**, not ECR — so the versioned images that do exist are on the registry the docs don't point at. Worth aligning whichever way the team prefers.

No action needed on this PR. Happy to open a fresh one in the ECR direction if that's useful.

PR #293, opened 2026-07-23
Add legal-judgments case study (reproduce an AGE + pgvector demo on Samyama)

Raghuvarma905:
## What this adds

A new **`legal-judgments`** case study that reproduces a public reference demo (PostgreSQL + Apache AGE + pgvector, by an OSS community member) on Samyama — following the existing `football` / `cricket` pattern.

**Files:**
- `examples/legal_judgments_loader.rs` + `examples/legal_judgments_common/mod.rs` — Rust loader (SDK)
- `case_studies/legal-judgments/` — `case.env`, `demo.py`, `queries.cypher`, `README.md`, `run.sh`, `demo.gif`
- one catalogue row in `case_studies/README.md`

## The graph

**589 Indian Supreme Court judgments (2016)**, public dataset, **CC-BY-4.0** → **4,462 nodes / 8,363 edges**.
The cited `section` is stored on the `CITES` edge, so section-level questions are answerable.

| Node | Count | Edge | Count |
|---|---|---|---|
| Topic 2,291 · Party 1,102 · Case 589 · Act 446 · Judge 34 | 4,462 | `ABOUT` 3,041 · `CITES{section}` 2,749 · `PARTY_IN{role}` 1,309 · `DECIDED` 1,264 | 8,363 |

## Verification

- **All 6 showcase queries pass the Definition-of-Done gate** (build fails if any returns zero rows).
- Reproduces the reference demo's **published results exactly**: top judge **Dipak Misra — 104**, **IPC §302 — 57**, **Constitution Article 32 — 36**.
- Validated the load against every dataset-card statistic (589 judgments, 34 judges, 3,041 topic labels, all 12 monthly counts, top acts) — all exact.

PR #294, opened 2026-07-23
  legal-judgments: add Apache AGE head-to-head benchmark to case study README

Raghuvarma905:
## Description 📝

Adds a **Benchmark — head-to-head vs Apache AGE** section to the `legal-judgments`
case study README (the case study itself was merged in #293).

The same 4,462-node graph was loaded into **both** Samyama and Apache AGE
(`apache/age`, AGE 1.7.0 on PostgreSQL 18.1) on the same machine, and the same
queries run against each — median of 40 warm round-trips (Apache AGE via a
persistent psycopg2 connection; Samyama via HTTP):

| Query | Apache AGE | Samyama | Samyama is |
|---|---|---|---|
| Judges by case count | 34 ms | 0.94 ms | 36× faster |
| Most-cited sections | 23 ms | 1.5 ms | 15× faster |
| Laws by topic breadth (2-hop) | 155 ms | 19 ms | 8× faster |
| Co-sitting judges (2-hop) | 25 ms | 16 ms | 1.5× faster |

**Why:** Apache AGE runs Cypher inside PostgreSQL via a `cypher()` function, so
Postgres re-plans every call (`EXPLAIN ANALYZE`: planning 8.7 ms + execution
5.5 ms). Samyama caches the plan and runs on a native in-memory graph. The gap is
largest on aggregation (15–36×) and narrows to 1.5× on the 2-hop join.

A **method note** is included ("same host, warm connection, median of 40; two
client transports; a like-for-like local comparison at this dataset size — not a
large-scale benchmark") so the numbers are reproducible and honestly scoped.

Also converts the Showcase table to **Query / Result / Gate**, so per-query timings
live only in the benchmark table (no double-timing).

### DIFF
--- a/case_studies/legal-judgments/README.md
+++ b/case_studies/legal-judgments/README.md
@@ -40,15 +40,46 @@ The `section` lives on the `CITES` edge, so section-level questions
 ("how many judgments cite IPC §302?") are answerable — reproducing the reference's
 headline result exactly.
 
+## Benchmark — head-to-head vs Apache AGE
+
+The same 4,462-node graph was loaded into **both** Samyama and Apache AGE (`apache/age`, AGE 1.7.0 on
+PostgreSQL 18.1) on the same machine, and the same queries run against each — median of 40 warm
+round-trips (Apache AGE via a persistent psycopg2 connection; Samyama via HTTP):
+
+| Query | Apache AGE | Samyama | Samyama is |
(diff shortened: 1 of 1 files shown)


nareshkolimi (approved):
Looks good to me

sandeepkunkunuru (approved):

PR #315, opened 2026-07-27
case study: vector-search benchmark vs pgvector (honest, scale-dependent)

Raghuvarma905:
## What

Adds a **vector-search benchmark vs pgvector** to the legal-judgments case study, completing the "Postgres (AGE **and pgvector**)" comparison requested by Shreyas Rao. The AGE (graph/Cypher) benchmark already lives in this case study; this adds the vector half.

## Results (real measurements)

Same synthetic vectors in both engines, at the repo's standard `vector_benchmark` dimensions (**128, 768**), HNSW + cosine, k=10, median of 40 warm queries.

| Dim | Vectors | pgvector (server compute) | pgvector (client, TCP) | Samyama (embedded) | vs server | vs client |
|---|---|---|---|---|---|---|
| 128 | 589 | 0.135 ms | 0.561 ms | **0.090 ms** | 1.5× | 6.2× |
| 128 | 10,000 | 0.869 ms | 1.557 ms | **0.253 ms** | 3.4× | 6.2× |
| 768 | 589 | 1.244 ms | 2.166 ms | **0.473 ms** | 2.6× | 4.6× |
| 768 | 10,000 | 1.867 ms | 2.978 ms | **0.993 ms** | 1.9× | 3.0× |

## Honest framing (deliberate)

- **Samyama is faster on pure search compute in every case** (1.5×–3.4×); the gap narrows at dim 768 where pgvector's SIMD compute is strongest.
- **3–6× real-world** because pgvector is reached over the network on every query; Samyama can run embedded (in-process).
- **Single engine** for graph + vectors vs Postgres + AGE + pgvector.

## Method / caveats (in the README)

- Synthetic vectors at dims 128 & 768 — the **same methodology and dimension set as Samyama's own `benches/vector_benchmark.rs`**. Measures k-NN **latency, not embedding quality**; vectors are not real judgment embeddings.
- pgvector server-side time isolated via `EXPLAIN ANALYZE`; Samyama via embedded SDK (the running v1.0.0 engine predates the HTTP `vector-index/rebuild` route, so vectors are measured embedded — the same in-process model as Samyama's published LDBC results).

### DIFF
--- a/case_studies/legal-judgments/README.md
+++ b/case_studies/legal-judgments/README.md
@@ -89,6 +89,36 @@ HTTP). A like-for-like local comparison at this dataset size — not a large-sca
   [graph.samyama.cloud](https://graph.samyama.cloud) (see above) — the sub-millisecond reads are exactly
   what make that force-directed exploration feel instant.
 
+## Benchmark — vector search vs pgvector
+
+The reference stack uses **pgvector** for semantic search. We loaded the *same* vectors
+(dimension 384, HNSW + cosine) into both **PostgreSQL 17 + pgvector** and **Samyama**, and ran
+identical k-NN queries (k = 10, median of 40 warm queries, multiple runs):
+
+| Corpus | pgvector (server compute) | pgvector (client, over TCP) | Samyama (embedded) |
+|---|---|---|---|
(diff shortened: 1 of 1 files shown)


nareshkolimi (approved):
Looks Good to me

Issue #455, opened 2026-08-15
Solver tests are unseeded and one flakes at ~0.05%: test_hs_sphere failed a full-workspace run


sandeepkunkunuru:
## What happened

A full `cargo test --release --workspace` run failed on:

```
thread 'test_hs_sphere' panicked at crates/samyama-optimization/tests/test_solvers.rs:206:5
```

It then passed **100/100** on re-run (40 in isolation, 60 as the full `test_solvers` binary), which is exactly the signature of a rare stochastic flake rather than a regression.

## Measured, not guessed

Every solver takes its randomness from `thread_rng()` — there is **no seed parameter anywhere in the crate**. So each of the 27 threshold assertions in `test_solvers.rs` is a probabilistic claim with an unmeasured failure rate.

`HSSolver` on the test's own problem and config (Sphere 2D, pop 50, 500 iterations), **2000 runs**:

| | fitness |
|---|---|
| min | 0.000001 |
| p50 | 0.000940 |
| p90 | 0.029934 |
| p99 | 0.212509 |
| p99.9 | 0.539017 |
| max | **2.259323** |
| `>= 1.0` (the assert threshold) | **1 / 2000 = 0.05%** |

So the threshold is not marginal in the typical case — the median is a thousand times better than it. The distribution simply has a long right tail that crosses 1.0 about once every two thousand runs.

Sweeping all 24 single-objective solvers (400 runs each, same problem/config) shows this is confined to two of them:

| Solver | threshold | p50 | p99 | max |
|---|---|---|---|---|
| **GSA** | 1.0 | 0.010264 | 0.293239 | **0.561977** |
| **HS** | 1.0 | 0.000845 | 0.144185 | **0.322919** |
| Firefly | 1.0 | 0.001407 | 0.008001 | 0.012160 |
| SA | 1.0 | 0.002326 | 0.013110 | 0.018957 |
| Jaya, QOJaya, ITLBO, TLBO, PSO, DE, GWO, GOTLBO, ABC, FPA, BMWR, SAMPJaya, EHRJaya, QORao | 0.01–0.1 | 0.000000 | 0.000000 | 0.000000 |
| BMR, BWR, Cuckoo, GA, Bat | 0.1–1.0 | ~0.000001 | ≤0.000039 | ≤0.000071 |

Twenty-two solvers have roughly a thousandfold margin. **HS and GSA are the only two whose tails come anywhere near their thresholds**, and HS is the one that empirically crosses.

[...]

PR #456, opened 2026-08-15
Set HS and GSA test thresholds from their measured tails

sandeepkunkunuru:
Refs #455.

A full `cargo test --release --workspace` run failed on `test_hs_sphere`. It then passed **100/100** on re-run (40 isolated, 60 as the full binary) — a rare stochastic flake, not a regression.

## Why it flakes

Every solver in the crate draws from `thread_rng()` and **nothing accepts a seed**, so all 27 threshold assertions in `test_solvers.rs` are probabilistic claims with unmeasured failure rates. Measuring the two whose tails looked close, at the tests' own problem and config (Sphere 2D, pop 50, 500 iterations, 2000 runs each):

| Solver | p50 | p99 | p99.9 | max | runs ≥ 1.0 |
|---|---|---|---|---|---|
| **HS** | 0.00094 | 0.213 | 0.539 | **2.259** | 1 / 2000 |
| **GSA** | 0.011 | — | 0.669 | **1.025** | 1 / 2000 |

Both cross their `< 1.0` threshold at roughly one run in two thousand. A sweep of all 24 single-objective solvers confirmed these two are the only ones at risk — the other 22 finish three orders of magnitude inside their thresholds (most at exactly 0.000000).

Note GSA was **not** the test that failed; it was found by measuring rather than by waiting for it to bite.

## The change

Both thresholds go to `5.0`, with the measured distribution recorded in a comment beside each assertion.

This is chosen, not rounded up arbitrarily: 5.0 clears the worst observed value (2.259) while leaving a **5000×** margin against HS's median and **450×** against GSA's. A solver that genuinely stopped converging moves by orders of magnitude and still fails loudly.

## What this does not do

It does not make the crate reproducible. `thread_rng()` means no run of this optimizer can be re-derived — by a test or by a user — which for a crate backing published Rao-family results is a defect beyond CI hygiene. That needs a seedable `SolverConfig`, tracked in #455, including the part that actually needs care: `rao.rs`, `de.rs` and `gotlbo.rs` call `thread_rng()` inside rayon closures, where seeding only the outer RNG would yield results that look reproducible and are not.

`test_solvers`: **36 passed, 0 failed.**

### DIFF
--- a/crates/samyama-optimization/tests/test_solvers.rs
+++ b/crates/samyama-optimization/tests/test_solvers.rs
@@ -193,7 +193,12 @@ fn test_gsa_sphere() {
     let solver = GSASolver::new(config);
     let result = solver.solve(&problem);
 
-    assert!(result.best_fitness < 1.0, "GSA failed: fitness {}", result.best_fitness);
+    // Threshold set from the measured distribution, not picked by eye (#455).
+    // GSA over 2000 runs on this problem/config: p50 0.011, p99.9 0.669, max 1.025
+    // -- it crosses 1.0 about once in 2000 runs, so `< 1.0` is a ~0.05% flake.
+    // 5.0 clears the observed tail while leaving a ~450x margin against the
+    // median, so a solver that genuinely stopped converging still fails here.
+    assert!(result.best_fitness < 5.0, "GSA failed: fitness {}", result.best_fitness);
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


sandeepkunkunuru:
## Why the threshold tests are staying unseeded

#473 landed, so `test_solvers.rs` *could* now pin a seed and tighten the thresholds back from 5.0. I looked at doing that and decided against it, which is worth recording rather than leaving as an omission.

A seeded threshold test checks **one trajectory**. It is deterministic and it will never flake — but a solver regression that shows up on most seeds and not the pinned one passes silently. That trades a 0.05%-per-run false alarm for an unbounded false negative, on tests whose entire job is to notice when a solver stops converging.

The unseeded test samples the distribution every CI run. With the thresholds set from the measured tails (HS max 2.259, GSA max 1.025 over 2000 runs each, both against a 5.0 threshold) the false-alarm rate is now far below the 1-in-2000 that started this, while the sampling property is kept.

The seeding from #473 is doing its real work in `tests/test_reproducibility.rs`, where determinism is the property under test rather than a means of quieting one.

Worth revisiting if these ever flake again: the answer then is a *third* test that pins a seed and asserts an exact expected fitness, added alongside the distribution tests rather than replacing them.

Issue #480, opened 2026-08-16
[CH-REGRESS] There is no CI running the test suite — 2571 tests gate nothing


sandeepkunkunuru:
## The gap

`.github/workflows/` contains three workflows:

| Workflow | Trigger | Runs tests? |
|---|---|---|
| `gpu-ci.yml` | `workflow_dispatch` **only** | yes, but manual — and its own header records that the self-hosted GPU runner is offline, so push/PR triggers were removed after they queued forever |
| `publish-crate.yml` | release | no |
| `publish-docker.yml` | release | no |

**No workflow runs `cargo test` on a pull request or a push to main.** The suite is 2571 tests and it gates nothing. Recent Actions history is `GPU CI | pull_request | cancelled` repeatedly.

In practice every merge is gated by whoever happened to run the tests by hand. That worked today — 17 PRs, each verified locally and on a clean host — and it is not a control, because it depends on the person and leaves no record on the PR.

## Why this is the foundational one

A large number of spec requirements are phrased as *"gated in CI"* and all of them presuppose a CI that runs anything:

- `PERF-18` — CI fails on >10% regression of any tracked query
- `LANG-01` / #436 — TCK pass rate may never decrease
- `CH-IMPORT`, `CH-API`, `CH-DOC-EXEC`, `CH-ALGO-PARITY` — spec 18 lists these as **commit-blocking** gates
- Spec 18's operating rule 1: *"Gates block releases, not commits — except [seven gates] which block commits because they are fast and catch the most damaging classes"*

None of those can exist yet. This issue is the prerequisite for all of them.

## What good looks like

A full sweep (`scripts/verify-sweep.sh`) takes ~50 minutes — tests, all examples, 12 benches, 11 case studies. That is right for a nightly and wrong for a PR. So it needs tiers.

**Tier 1 — PR lane, target ≤10 min**
- [ ] `cargo test --workspace --release` on ubuntu-latest, on `pull_request` and `push: main`
- [ ] `cargo clippy -- -D warnings` and `cargo fmt --check`
- [ ] System deps installed explicitly (`clang`, `libclang-dev` — `zstd-sys` is the standard first-build failure on a clean host, already documented in CONTRIBUTING)
- [ ] Cargo cache keyed on `Cargo.lock` so this is not a 40-minute cold build every run
- [ ] **Required status check on `main`** — a green tick nobody has to enforce by hand

**Tier 2 — nightly**
[...]

PR #488, opened 2026-08-16
DO NOT MERGE: proving the CI gate fails a broken test

sandeepkunkunuru:
Not for merge. Exists to satisfy the last item of #480's definition of done: demonstrate that a deliberately broken test actually fails the pull-request check.

An untested CI config is not a control. This adds one failing assertion, waits for the check to go red, and is then closed and deleted.

Expected: the `test (ubuntu-latest)` check fails on `deliberately_broken_to_prove_ci_catches_it`.


sandeepkunkunuru:
Confirmed: the check went red on `deliberately_broken_to_prove_ci_catches_it` with `assertion `left == right` failed`, so the gate added in #486 does fail a pull request rather than merely reporting green.

That was the last open item on #480's definition of done that could be satisfied without repo-admin rights. Closing and deleting the branch.

PR #548, opened 2026-08-17
docs: make the README front screen match what we can verify

HimadriSur:
The front screen makes three claims we can't currently back, while the two most
credible artifacts in the repo sit below a 90-line Compose block. Four lines, no
content restructuring.

### 1. The cost headline

```diff
- The graph database that queried 1 billion edges for $2.50
+ Graph traversal, OpenCypher, and vector search in one Rust engine.
```

`docs/BENCHMARKS.md` contains no cost figure — I searched it for `$2`, `2.50`, `cost`
and `1 billion`, and there is no methodology, hardware note or provenance for this
anywhere in the repo. It's the most prominent claim on the page and the only one with
nothing behind it.

That reads worse now than it did a month ago, because #501/#502 set a much higher bar
in the same repo:

> "The SF10 column is still unverified and **should not be quoted**."
>
> "The old numbers were not fabricated, they were merely unprovable."
>
> "The previously documented 3,181,724 / 17,256,038 matched neither, and was 17% and
> 22% low. Anything divided by those counts was wrong by the same margin."

A reader who follows that standard into `BENCHMARKS.md` and then looks back at line 5
learns the wrong thing about how carefully this project handles numbers. The
replacement is descriptive, so it can't go stale.

The subtitle above it already positions the project, so nothing is lost. **Note the
repo *description* carries the same sentence** — that's a settings change and needs
someone with admin.

### 2. Version badge was hardcoded to 1.1.0

```diff
- img.shields.io/badge/version-1.1.0-blue
+ img.shields.io/github/v/release/samyama-ai/samyama-graph?color=blue&label=version
```

v1.7.0 shipped on 21 July; the badge has said 1.1.0 since. Now reads the latest release
directly, so it can't drift again.

### 3. Tests badge was a hardcoded count

```diff
- img.shields.io/badge/tests-2238_passing-brightgreen
+ actions/workflows/ci.yml/badge.svg?branch=main
```

`tests-2238_passing` is a static string that was true whenever it was typed. Since #486
added a lane that actually runs the tests, a live CI badge is both honest and more
useful — and a frozen count sitting next to real CI is the kind of detail that makes a
reader discount every other badge.

### 4. Surfaced the two artifacts worth finding

One line above the Quickstart pointing at `case_studies/` and `docs/BENCHMARKS.md`.
Someone landing here can currently read 90 lines of Docker Compose without learning that
11 real knowledge graphs run in one command, with a gate that fails the build if any
showcase query returns zero rows.

---

Happy to split this into separate PRs, or to drop any item — item 1 is the one I'd argue
for keeping.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,11 @@
   <h1 align="center">Samyama Graph</h1>
   <p align="center">A Rust-native graph-vector database for GraphRAG, knowledge graphs, and billion-edge analytics.</p>
   <p align="center">
-    <strong>The graph database that queried 1 billion edges for $2.50</strong>
+    <strong>Graph traversal, OpenCypher, and vector search in one Rust engine.</strong>
   </p>
   <p align="center">
-    <a href="https://github.com/samyama-ai/samyama-graph/releases"><img src="https://img.shields.io/badge/version-1.1.0-blue" alt="Version"></a>
-    <a href="https://github.com/samyama-ai/samyama-graph/actions"><img src="https://img.shields.io/badge/tests-2238_passing-brightgreen" alt="Tests"></a>
+    <a href="https://github.com/samyama-ai/samyama-graph/releases"><img src="https://img.shields.io/github/v/release/samyama-ai/samyama-graph?color=blue&label=version" alt="Version"></a>
+    <a href="https://github.com/samyama-ai/samyama-graph/actions/workflows/ci.yml"><img src="https://github.com/samyama-ai/samyama-graph/actions/workflows/ci.yml/badge.svg?branch=main" alt="CI"></a>
     <a href="LICENSE"><img src="https://img.shields.io/badge/license-Apache_2.0-blue" alt="License"></a>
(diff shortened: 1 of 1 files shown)


nareshkolimi (approved):
Looks good to me

sandeepkunkunuru:
Superseded by #1291 (`79cd939`), which makes the change you asked for: the $2.50 headline is gone, replaced by claims the harness re-measures — TCK 3,845 of 3,847 evaluated scenarios, SNB Interactive 21/21 and BI 20/20 with no timeouts, 1B edges on one machine. The scale run keeps its row and now states what the cost figure actually is: about $2.50 of spot time for one run on one instance type, not a price list.

Your reasoning was right and sat open for a month. The badges you did not touch were also two releases stale (1.1.0, 2,238 tests) and are now 1.8.0 and 4,494.

Issue #545, opened 2026-08-17
[PERF-10] set_node_property writes every value twice — 808 MB of duplication on LDBC SF1


sandeepkunkunuru:
## Measured

Same dataset, same machine, the only difference being which storage path the loader uses (#544):

| | row storage only | row **and** columnar |
|---|---:|---:|
| load time | 50.4 s | 59.4 s (+18%) |
| peak RSS | 10.70 GB | **11.50 GB** (+7.6%) |

**+808 MB on 3,727,429 nodes — about 217 bytes per node**, for values that are already stored elsewhere.

## Why it is duplication rather than two useful copies

`GraphStore::set_node_property` writes to both:

```rust
self.node_columns.set_property(idx, &key_str, val.clone());   // columnar
…
versions.push(node);                                          // row, inside the Node
```

and `Value::resolve_property` reads the column first, falling back to the row only when the column returns `Null`:

```rust
Value::NodeRef(id) => {
    let prop = store.node_columns.get_property(id.as_u64() as usize, property);
    if !prop.is_null() { prop }
    else if let Some(node) = store.get_node(*id) { … }
}
```

So on any graph built through the API — `CREATE`, CSV import, the LDBC loader after #544 — **the row copy is never read**. It is written, it is carried in the MVCC version chain, it is exported, and nothing consults it.

This matters more after #535: a columnar read is now 12 ns against row storage's 61.7 ns, so the column is not merely the preferred path, it is 5× the faster one.

## Why it is not simply removable

Three things depend on the row copy existing, and each needs an answer before it goes:

1. **Types the columnar store cannot hold.** `ColumnStore::set_property` silently drops anything that is not `Integer`, `Float`, `String` or `Boolean` — `Array`, `Map`, `DateTime`, vectors. Those *only* live in the row today, which is why the fallback exists and why removing it naively would lose them.
2. **MVCC.** Node versions carry a `PropertyMap` each; the columnar store holds only the current value. Historical reads (`get_node_at_version`) go through the row.
3. **Snapshot export** walks `all_nodes()` and reads `node.properties`. It would need to read the merged view instead — the same defect class as #303 and #507, where a rebuild was forgotten after an import.

## Options

[...]

Issue #364, opened 2026-08-14
DETACH DELETE does not clear property columns — new nodes resurrect deleted values


sandeepkunkunuru:
## Summary

`DETACH DELETE` does not clear the columnar property store. A node created afterwards with the same label **inherits the previous generation's value** for any property column it does not itself set.

This is a data-integrity bug: values from deleted data reappear on new data.

## Reproduction (self-contained)

```cypher
CREATE (:GhostProp {id:"a", ghost:"LEAKED"});
MATCH (n:GhostProp) DETACH DELETE n;
MATCH (n:GhostProp) RETURN count(n) AS n;     -- 0, as expected

CREATE (:GhostProp {id:"b"});                 -- note: no `ghost` property
MATCH (n:GhostProp) RETURN n.id, n.ghost;
-- | "b" | "LEAKED" |     WRONG, expected NULL
```

A global wipe does not help either:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:GhostProp {id:"c"});
MATCH (n:GhostProp) RETURN n.id, n.ghost;
-- | "c" | "LEAKED" |     still wrong
```

## How we hit it

An internal field leaked into `Sensor` node properties from our loader. We fixed the loader, regenerated the source data (verified clean on disk), reset the graph with `DETACH DELETE` and reloaded — and **every `Sensor` still reported the stale value**. It looked for a while like the loader fix had failed.

The only thing that cleared it was stopping the server and starting it against a fresh data directory.

## Impact

- `DETACH DELETE` is not a usable reset between experiments whenever the node's property *schema* changes (adding values is fine; removing a property is not).
- Any ETL that reloads a graph after dropping a property will silently serve the old values.
- Test suites that reset state between cases can leak state across cases.

## Workaround

Treat a fresh data directory as the only true reset:

```bash
pkill -f '[t]arget/release/samyama'
rm -rf <data-dir>/samyama_data
samyama --http-port 8080
```

## Related smaller issue

`<>` against a null property matches, where openCypher would evaluate `null <> ""` as null and filter the row out:

```cypher
MATCH (s:Sensor) WHERE s.missing_prop <> "" RETURN count(s.id)   -- matches all rows
```

This made the bug above harder to diagnose, since the natural "is it still there?" check reported rows for nodes whose value was actually null.

## Environment

- Samyama Graph **v1.7.0** (OSS)

Issue #559, opened 2026-08-17
[PERF-05] Filter goes parallel on batch size, not predicate cost — and loses 1.4-1.8x on every predicate a real query has


sandeepkunkunuru:
## Measured

`FilterOperator::next_batch` switches to a rayon parallel filter when a batch has **256 or more records**. Batch size is not a proxy for how much work a predicate does, and the threshold is reached by essentially every batch — the batch size is 65,536.

Interleaved A/B in one process, three rounds, minimum taken, 1,000,000 rows through the filter. Only the threshold varies:

| predicate | parallel | sequential | |
|---|---:|---:|---|
| *(no filter — baseline)* | 178.1 ms | 180.5 ms | tie |
| `WHERE 1 = 1` | 194.2 ms | 191.0 ms | tie |
| `WHERE i.v > 500` | 404.8 ms | **252.0 ms** | **sequential 1.61×** |
| `WHERE i.v > 500 AND i.w > 5` | 486.7 ms | **327.4 ms** | **sequential 1.49×** |
| `WHERE i.name CONTAINS "99"` | 525.7 ms | **285.8 ms** | **sequential 1.84×** |
| `WHERE toUpper(i.name) CONTAINS "99"` | 512.1 ms | **363.8 ms** | **sequential 1.41×** |
| 4 conjuncts, 2 of them string ops | **461.5 ms** | 604.9 ms | parallel 1.31× |

**Parallel wins exactly once**, on the heaviest predicate tested. Every predicate that looks like something a real query writes loses, and the simplest one loses hardest.

## Why

A `Record` is a `Vec<(Arc<str>, Value)>`. Sending records through `into_par_iter().filter(...).collect()`:

- **churns `Arc` refcounts across cores.** Every clone and drop of an `Arc<str>` binding name is an atomic RMW on a cache line shared by every thread, and the variable names are shared by every record. That is the textbook contention pattern, and it scales with core count in the wrong direction.
- **reallocates on collect.** A parallel collect builds per-thread vectors and merges them, moving every surviving `Record`.

Against that, a predicate like `i.v > 500` is one property read and one integer comparison — around 80 ns. There is nothing to amortise the coordination against.

The `constant true` row is the control: it does no memory access and no refcount work beyond the record move, and the two paths tie. The moment the predicate touches a property, parallel loses.

## The fix

[...]

PR #560, opened 2026-08-17
Locate a property's column once per query, not once per row

### LINKED ISSUE #557: [PERF-05] Reading one property costs ~115 ns per row — the name is hashed, and the column looked up, once per row
Split out of #521, with the number it took to find it.

## Measured

One host, one fixture, `PROFILE`, **`Aggregate` self time** so the traversal around it is not folded in. 1,000,000 input rows over 50,000 groups:

| query | Aggregate self | ns per input row |
|---|---:|---:|
| `RETURN count(i)` — no grouping, no property read | 65.2 ms | **65** |
| `RETURN f.fid, f.title, sum(i.v)` — one property read per row | 180.4 ms | **180** |

Same operator, same rows, same fold. The difference is **one property resolution per row: ~115 ns.**

For scale: the `Expand` that produced those million rows — following adjacency, allocating a record per row — cost 145 ms, or 145 ns/row. **Reading one integer property costs about as much as producing the row it sits on.**

## Where it goes

`Value::resolve_property` → `ColumnStore::get_property`:

```rust
pub fn get_property(&self, idx: usize, key: &str) -> PropertyValue {
    self.columns.get(key).map(|col| col.get(idx)).unwrap_or(PropertyValue::Null)
}
```

Per row, for a property whose name and column are fixed for the whole query:

1. hash the property name string and probe `FxHashMap<String, Column>`;
2. `col.get(idx)` — an array index on a dense column, a second hash probe on a sparse one;
3. if that yields `Null`, fall back to the per-node row `HashMap` (#545), which is a pointer chase into a separate allocation per node.

Step 1 is loop-invariant. `get_column(key) -> Option<&Column>` already exists and is commented "optimized batch read for a single property" — **nothing in the executor calls it.**

Step 3 is only free when the column hits. For a property that lives *only* in row storage, every row pays a failed hash probe and then a cold pointer chase.

## Why it matters

This is not an aggregation cost, so #521 could only ever move a fraction of it. It is paid by `Filter`, `Project`, `Sort`, `Aggregate` and every expression evaluation — everything that touches `x.prop`, which is most of every query.

`PERF-05` and `SLT-2`. It also caps what #545 and #491 can be worth: removing the duplicate row copy helps memory, but the per-row lookup stays until the lookup itself is hoisted.

## What to try, cheapest first

- [ ] **Resolve the column once per operator, not once per row.** The property name is fixed at plan time. An operator holding `Option<&Column>` (or a column id) turns step 1 into nothing. The obstacle is lifetimes — `store` arrives per call, not per operator — so this probably wants a `ColumnId` resolved at plan time and an index-based lookup rather than a borrowed reference.
- [ ] **Intern property names to a `u32` at parse time**, so `columns` is a `Vec<Column>` indexed directly. Bigger change; also removes the `String` keys from the store.
- [ ] Measure the row-storage fallback separately — a property present in the column and one present only in the row map should be timed side by side, because the second is the pathological case and nothing currently tells them apart.

## Definition of done

- [ ] A microbenchmark reporting ns per property read, split by dense column / sparse column / row-storage-only
- [ ] The hoist implemented for at least `Filter` and `Aggregate`
- [ ] Re-profile the fixture above and record the new Aggregate self time against the 65 ns/row floor
- [ ] LDBC SF1 re-run, since IC5 and IC3 are dominated by exactly this

## Reproducing

`benches/aggregate_grouping.rs` (added in #521's PR) builds the fixture. The `1 key int, + sum` and `no grouping (floor)` rows differ by one property read per row.

sandeepkunkunuru:
Closes #557.

## Measured

Isolating the read as `sum(i.v)` minus `count(i)` over the same 1,000,000 rows, so only the property access differs:

| | main | this branch |
|---|---:|---:|
| one property read, per row | 23.98 ns | **16.90 ns** (−30%) |

And directly, in `benches/property_access.rs`:

| access | main | this branch |
|---|---:|---:|
| dense int column, **id order** | 19.5 ns | **17.1 ns** |
| dense int column, **scattered order** | 37.6 ns | **22.9 ns** (−39%) |

## What changed

`get_property(idx, "title")` hashed the property name against the column index **on every call**. An operator looping over a million rows evaluating the same expression did that a million times to reach the same column — and the name is fixed at plan time.

Columns now have **stable ids**. `ColumnStore` holds a `Vec<Column>` with a name→id map beside it; a caller resolves the id once and reads by index. Ids are safe to cache because columns are only ever appended — `clear_row` clears a row's *values*, and nothing removes a column.

`PropertyCursor` is the caller. It memoises the column id on first read and is otherwise **indistinguishable** from `Value::resolve_property` — which is the point, and what `tests/property_cursor.rs` asserts by running both paths and comparing.

All four aggregate fold paths now read through cursors, for group keys and aggregate arguments alike.

## Two things it deliberately does not do

- **It does not cache a miss.** A later `SET` may create the column, so a cursor that remembered "no column" would keep returning null. There is a test.
- **It keeps the row-storage fallback.** A value whose type has no typed column representation is readable only from the per-node map (#545); dropping that path would silently return null for complex types. There is a test.

## The honest frame

The benchmark gained a **scattered-order** case, because the gap between access *orders* turned out larger than the gap this change closes:

```
dense int, id order        17.1 ns
dense int, scattered       22.9 ns
a bare Vec<i64>, id order   0.6 ns
a bare Vec<i64>, scattered  3.1 ns
```

Most of a scattered read is a DRAM round trip that no representation avoids. That matters because a scan reads in id order but **anything downstream of an `Expand` does not** — it arrives at nodes in adjacency order. Saying so is the right frame for what is left.

Also visible and worth its own work: **row storage costs 150 ns** against a column's 17 — 9×. That is #545.

## Testing

11 tests in `tests/property_cursor.rs`, all differential — same input, both paths, same answer. Covering: a column-backed read; an absent property; a column created *after* the cursor first missed; a value living only in row storage; a row with no value in a column that exists; an edge property; an unbound variable; a non-graph value (`datetime.year`); id stability across 64 appended columns; a recycled node id (#364); and a whole-graph fold against a hand-computed sum.

Suite on a dedicated box: **exit 0, 2,746 tests** (2,735 before, +11). LDBC SF1 unchanged at **21/21 in 24.2 s**.

### DIFF
--- a/benches/property_access.rs
+++ b/benches/property_access.rs
@@ -114,6 +114,25 @@ fn main() {
     let columns = &fx.store.node_columns;
     let ids = &fx.rows;
 
+    // The same rows, visited in scattered order. A scan reads in id order, but
+    // anything downstream of an `Expand` does not: a traversal arrives at nodes
+    // in adjacency order, which is scattered with respect to the id space. Since
+    // the two orders read exactly the same values through exactly the same code,
+    // any difference between them is cache and nothing else.
+    let scattered: Vec<usize> = {
+        let mut v = ids.clone();
+        // Deterministic shuffle -- a benchmark that varies run to run cannot be
+        // compared across commits (#529).
+        let mut state = 0x2545_F491_4F6C_DD1Du64;
+        for i in (1..v.len()).rev() {
+            state ^= state << 13;
+            state ^= state >> 7;
+            state ^= state << 17;
+            v.swap(i, (state % (i as u64 + 1)) as usize);
+        }
+        v
+    };
+
     println!("{:<46} {:>10}", "access", "ns");
     println!("{:-<46} {:->10}", "", "");
 
@@ -173,6 +192,39 @@ fn main() {
         acc
     });
 
+    let scattered_ns = time("get_property, dense int, scattered order", rows, || {
+        let mut acc = 0u64;
+        for &idx in &scattered {
+            if let PropertyValue::Integer(v) = columns.get_property(idx, "dense_int") {
+                acc = acc.wrapping_add(v as u64);
+            }
+        }
+        acc
+    });
+
+    let scattered_hoisted = time("  ... with the column hoisted", rows, || {
+        let mut acc = 0u64;
+        if let Some(col) = columns.get_column("dense_int") {
+            for &idx in &scattered {
+                if let PropertyValue::Integer(v) = col.get(idx) {
+                    acc = acc.wrapping_add(v as u64);
+                }
+            }
+        }
+        acc
+    });
+
+    let scattered_vec = {
+        let plain: Vec<i64> = (0..rows as i64).collect();
+        time("a dense Vec<i64>, scattered order", rows, || {
+            let mut acc = 0u64;
+            for &idx in &scattered {
+                acc = acc.wrapping_add(plain[idx.min(plain.len() - 1)] as u64);
+            }
+            acc
+        })
+    };
+
     let row_nodes = &fx.row_nodes;
     time("row storage, node.properties HashMap", rows, || {
         let mut acc = 0u64;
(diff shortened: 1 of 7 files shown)


sandeepkunkunuru:
Superseded, not abandoned: this branch's commits are already on `main`.

While the GitHub API was returning 503s this PR could not be merged, and the branch for #559 was based on it. When #561 merged, GitHub squashed **both** commits into f40dc06 — so the column-id and `PropertyCursor` work is in `main`, just under #559's commit message. Verified: `ColumnId`, `PropertyCursor` and `tests/property_cursor.rs` are all present on `main`.

Recorded in 16da8a7 so the history is not misleading.

Issue #549, opened 2026-08-17
uc4_optimizer_beats_uniform_baseline is stochastic and fails intermittently in a full-suite run


sandeepkunkunuru:
## What happened

`tests/uc4_kg_completion_test.rs::uc4_optimizer_beats_uniform_baseline` failed once during a `cargo test --release --workspace` run, and passed **12 out of 12** when run on its own immediately afterwards, on the same commit and the same machine.

```
thread 'uc4_optimizer_beats_uniform_baseline' panicked at tests/uc4_kg_completion_test.rs:215
```

Line 215 is:

```rust
assert!(
    best_fit >= 0.3,
    "expected Hits@5 ≥ 0.3 over 3 seeds; got {best_fit}"
);
```

## Why it is intermittent

The test runs `QOJayaSolver` — a stochastic metaheuristic — three times with `population_size: 20, max_iterations: 40`, keeps the best fitness, and requires it to clear a fixed 0.3 threshold. The comment above it already concedes the shape of the problem:

```rust
// and a single run on this small problem can fail to beat baseline.
```

Three seeds reduces the failure rate; it does not remove it. A full-suite run also puts the solver under CPU contention from every other test binary, which is the difference between the failing context and the passing one.

## Why it matters more than one flaky test

`PERF-18` / `CH-REGRESS` calls for a **blocking** nightly gate. A test that fails perhaps one run in twenty teaches people to re-run CI rather than read it, and once that habit exists a real regression gets the same treatment. #530's calibration test had the same problem for the same reason — an assertion whose truth depends on the machine — and was fixed by measuring something robust rather than by loosening the bound.

## Options

1. **Seed the solver deterministically.** `#473` made every solver seedable, including under rayon. If this test fixes the seed, the assertion becomes a property of the code rather than of the draw. Strongly preferred — it turns a probabilistic test into a real one.
2. **Assert the comparison, not the constant.** `best_fit >= baseline_fit` is the claim the test's name makes; `best_fit >= 0.3` is a separate, absolute claim that depends on the fixture and the budget. Splitting them would at least say which one broke.
3. **Raise the seed count.** Cheapest, weakest — it lowers the rate without bounding it.

I would take (1), and keep (2) as the assertion structure: with a fixed seed both bounds become deterministic and the failure message says which property was violated.

## Definition of done

- [ ] The solver is seeded, so the test is deterministic on a given commit
- [ ] The comparison against baseline and the absolute threshold are separate assertions
[...]

Issue #538, opened 2026-08-17
shortestPath anchored by WHERE id(a) = … is 1000x slower than the same query anchored inline


sandeepkunkunuru:
## Measured

4,000 nodes, out-degree 30, endpoints three hops apart. Same graph, same answer, same session:

| query | time | rows |
|---|---:|---:|
| `shortestPath((a:N {seq: 1})-[:KNOWS*]-(b:N {seq: 3500}))` | **4.2 ms** | 1 |
| `allShortestPaths((a:N {seq: 1})-[:KNOWS*]-(b:N {seq: 3500}))` | **4.8 ms** | 48 |
| `shortestPath((a:N)-[:KNOWS*]-(b:N)) WHERE id(a) = 1 AND id(b) = 3500` | **4,219 ms** | 1 |

**~1000×** for what is logically the same question.

## The plan says why

```
EXPLAIN MATCH p = shortestPath((a:N)-[:KNOWS*]-(b:N))
        WHERE id(a) = 1 AND id(b) = 3500 RETURN length(p) AS len

Project (length(p) AS len)
+- Filter (id(b) = Integer(300))
   +- Unknown
```

Two things:

1. **`id(a) = 1` has vanished** and `id(b) = 3500` sits *above* the path search. So the search is not anchored on either endpoint by the predicate — it runs and is filtered afterwards, which for a pairwise operator means running it across a cross product of candidate endpoints.
2. **The operator renders as `Unknown`**, because `ShortestPathOperator` did not implement `describe()`. Fixed in the same PR as #516; noting it here because it is why the plan could not be read before.

The inline form lowers each endpoint to an anchored scan, so the operator receives exactly one `(source, target)` pair.

## Why it matters

- `id()` is the most natural way to anchor a path query when you already hold node ids, and it is what an application does. Inline property syntax requires a property that happens to be indexed.
- The two forms are equivalent in openCypher. A user has no way to know that one of them is 1000× the other, and nothing in the plan told them until `describe()` existed.
- It is the same class as #328: a highly selective predicate that the planner does not use to choose an anchor.

## Not the cause of #516

Worth separating, because they look alike. #516 was the *algorithm* — `allShortestPaths` disabled its visited set and enumerated walks. That is fixed, and LDBC IC14 uses the inline form, so IC14's timeout was genuinely #516's. This issue is what remains once the algorithm is right.

## Suggested fix

[...]

PR #585, opened 2026-08-18
Pin both shortestPath endpoints, and stop asserting wall clocks in tests

### LINKED ISSUE #584: [PERF-05] shortestPath anchors only its start endpoint on id(), so the BFS runs once per node in the label
## Measured

Same graph, same answer, two ways of naming the endpoints:

```cypher
-- 329x slower
MATCH p = shortestPath((a:N)-[:KNOWS*]-(b:N)) WHERE id(a) = 1 AND id(b) = 6 RETURN length(p)

-- baseline
MATCH p = shortestPath((a:N {seq: 0})-[:KNOWS*]-(b:N {seq: 5})) RETURN length(p)
```

On a 1,200-node chain: **38.8 ms against 12.79 s** — 329×.

## The plans say why

```
-- id() form
Project (length(p) AS v)
+- Filter (id(a) = Integer(1))
   +- Filter (id(b) = Integer(6))
      +- ShortestPath ((a)-[:KNOWS*]-(b), Both)
         +- CartesianProduct
            +- NodeById (var=a, ids=[1])      ← pinned
            +- NodeScan (var=b, labels=["N"]) ← every node

-- inline form
      ShortestPath ((a)-[:KNOWS*]-(b), Both)
      +- CartesianProduct
         +- Filter (a.seq = Integer(0))
         |  +- NodeScan (var=a, ...)
         +- Filter (b.seq = Integer(5))       ← filtered *below* the join
            +- NodeScan (var=b, ...)
```

Only the **start** endpoint is anchored. `b` scans the whole label, so the cartesian product yields one pair per node, **the BFS runs once per node**, and the `Filter (id(b) = …)` above discards all but one result.

The inline form pushes both property filters below the join, so it runs one BFS.

## Where

The shortestPath branch of the planner builds its target endpoint as a plain scan and consults only the pattern's inline properties:

```rust
let target_scan: OperatorBox = Box::new(NodeScanOperator::new(
    target_var.clone(),
    last_segment.node.labels.clone(),
));
let target_op = if let Some(ref props) = last_segment.node.properties { … };
```

`find_id_predicate` — added in #538 and wired into the start-anchored builder and `build_path_from_anchor` — is never called for it. Neither is `find_index_predicate`, so an indexed property on the target endpoint is presumably in the same position.

## How it survived #538

#538 added a test named `both_endpoints_of_a_path_can_be_pinned_by_id`, and **it passed**. It asserted the query finished in under 500 ms on a 2,000-node chain in `--release`, which it did — while doing 2,000 BFS runs instead of one. The assertion was satisfied by a plan that was merely fast enough on that machine.

That PR's description claims the shape works. It does not, and this issue is the correction.

## Definition of done

- [x] The target endpoint takes an `id()` anchor
- [ ] A plan-shape assertion — two `NodeById`, no `NodeScan` — rather than a wall-clock one
- [ ] The same for `find_index_predicate` on the target endpoint, which looks equally absent
- [ ] Check `allShortestPaths`, which shares the branch

## Reproducing

`cargo test --test id_anchor`, which now compares the `id()` form against the inline form on the same graph rather than against a clock.

sandeepkunkunuru:
Closes #584. Fixes the red CI.

## Why CI was red

`--test id_anchor` failed two timing assertions I wrote in #538. They assert **milliseconds**, and **CI builds in debug** while I had verified everything in `--release`. An absolute bound encodes the machine *and the build profile* that wrote it — the same mistake #549 was about, made in the other direction, by me.

Both are now **ratios**, invariant to machine speed and build profile:

- the point lookup is timed on a 2,500-node graph and a 50,000-node graph and required to be **flat** — a scan would be 20×;
- the shortestPath form is timed against **the same query written with inline properties**, which is the comparison #538 was actually about.

## Rewriting the second one found a real bug

And the old assertion is what hid it. `shortestPath` anchored only its **start** endpoint:

```
Filter (id(a) = Integer(1))
  Filter (id(b) = Integer(6))
    ShortestPath ((a)-[:KNOWS*]-(b), Both)
      CartesianProduct
        NodeById (var=a, ids=[1])       ← pinned
        NodeScan (var=b, labels=["N"])  ← every node
```

So the BFS ran **once per node in the label**, and the filter above discarded all but one result: **12.79 s against the inline form's 38.8 ms** on a 1,200-node chain — **329×**.

The old test asserted "under 500 ms in release on 2,000 nodes". That plan satisfied it, while doing 2,000 BFS runs instead of one. #538's description claims this shape works; it did not, and this is the correction.

## What is fixed

- the target endpoint takes the same **`id()`** anchoring the start had;
- and the same **index** anchoring — `WHERE a.seq = 0 AND b.seq = 5` planned as `IndexScan(a) × NodeScan(b)`. That is the `WHERE` form of what the inline form `(b:N {seq: 5})` already got right, which is exactly why it went unnoticed;
- **`allShortestPaths`** shares the branch and is fixed with it — asserted, not assumed.

## Testing

Five new tests assert the **plan shape** — two `NodeById`, no `NodeScan` — because that is precisely what a timing assertion failed to pin. Plus the answers, and an endpoint left deliberately unconstrained, since a plan-shape change on a path operator is where a row goes missing quietly.

Verified on a dedicated box **in both profiles**, which is the check that was missing all along:

| | exit | tests | failed |
|---|---|---:|---:|
| debug (what CI runs) | 0 | 2,825 | 0 |
| release | 0 | 2,825 | 0 |

LDBC SF1: **21/21 in 14.2 s**.

### DIFF
--- a/src/query/executor/planner.rs
+++ b/src/query/executor/planner.rs
@@ -2428,11 +2428,31 @@ impl QueryPlanner {
                     .collect();
                 let all_paths = matches!(path.path_type, PathType::AllShortest);
 
-                // We need the target node to be scanned too — create a CartesianProduct with target scan
-                let target_scan: OperatorBox = Box::new(NodeScanOperator::new(
-                    target_var.clone(),
-                    last_segment.node.labels.clone(),
-                ));
+                // We need the target node to be scanned too — create a CartesianProduct with target scan.
+                //
+                // The target gets the same `id()` anchoring the start already
+                // got (#538), because without it only *one* endpoint was
+                // pinned: `WHERE id(a) = 1 AND id(b) = 6` planned as
+                // `NodeById(a) x NodeScan(b)`, so the BFS ran once per node in
+                // the label and a filter above discarded all but one result.
+                // Measured against the same query written with inline
+                // properties, that cost 329x (#584).
+                let target_id_pred = find_id_predicate(&target_var, &deferred_predicates);
+                let target_scan: OperatorBox = match &target_id_pred {
+                    Some((_, ids)) => {
+                        let mut op = NodeByIdOperator::new(ids.clone(), target_var.clone());
+                        if !last_segment.node.labels.is_empty() {
+                            // A scan by id bypasses the label index, so the
+                            // pattern's labels still have to be checked.
+                            op = op.with_labels(last_segment.node.labels.clone());
+                        }
+                        Box::new(op) as OperatorBox
+                    }
+                    None => Box::new(NodeScanOperator::new(
+                        target_var.clone(),
+                        last_segment.node.labels.clone(),
+                    )),
+                };
                 // Add property filter for target node
                 let target_op = if let Some(ref props) = last_segment.node.properties {
                     if !props.is_empty() {
--- a/tests/id_anchor.rs
+++ b/tests/id_anchor.rs
@@ -118,44 +118,141 @@ fn an_id_list_anchors_on_every_element() {
     assert_eq!(got, vec![2, 4, 6]);
 }
 
+/// Total time to run `cypher` against `store` `reps` times.
+///
+/// Repeated because a single anchored lookup is microseconds, and a ratio
+/// between two microsecond measurements is mostly timer noise.
+fn repeat(store: &GraphStore, cypher: &str, reps: u32) -> std::time::Duration {
+    let query = parse_query(cypher).expect("parse");
+    // Warm, so the first call does not pay for statistics the rest skip.
+    let _ = QueryExecutor::new(store).execute(&query).unwrap();
+    let started = std::time::Instant::now();
+    for _ in 0..reps {
+        let batch = QueryExecutor::new(store).execute(&query).unwrap();
+        std::hint::black_box(batch.records.len());
+    }
+    started.elapsed()
+}
+
 #[test]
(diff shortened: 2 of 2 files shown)


sandeepkunkunuru:
**Correction: this PR does not contain the index-anchoring half it describes.**

The description says "and the same index anchoring — `WHERE a.seq = 0 AND b.seq = 5` planned as `IndexScan(a) × NodeScan(b)`". That change was written and tested locally, but a **bash parse error killed the `git add` that would have staged it** — the shell failed to parse a heredoc later in the same command line, so nothing in the chain ran, silently. The commit that merged here carries the `id()` anchoring and the ratio tests, and nothing else.

The five tests named in the Testing section are likewise split: `both_shortest_path_endpoints_reach_a_node_by_id` is here; `an_indexed_property_pins_both_endpoints_too`, `all_shortest_paths_pins_both_endpoints_too`, `the_indexed_form_still_answers` and `an_unanchored_endpoint_still_works` are not.

What *is* accurate about this PR: the CI fix, the ratio assertions, the `id()` anchoring of the target endpoint, and the 329× measurement that motivated it. The both-profile verification (debug 2,825 / release 2,825) was run against exactly this state, so those numbers stand.

The remainder is in #586, with the same verification. Recorded here rather than quietly, because the same failure mode — a claim in a PR body that the tests did not actually check — is what let #584 hide inside #538 in the first place.

Issue #617, opened 2026-08-18
Cypher clauses can appear in any order; the grammar hard-codes one — the largest remaining TCK blocker


sandeepkunkunuru:
After the TCK sweep in #615 took the pass rate from 46.6% to 53.5%, **107 of the 176 remaining parse failures are one structural problem**: the grammar fixes the order in which clauses may appear, and Cypher does not.

## The shape

```
unwind_stmt = { unwind_clause+ ~ (match_clause+ ~ where?)* ~ (with_clause ~ …)* ~ create_clause? ~ merge_inline? ~ … ~ return_clause? }
```

`create_clause` sits **after** the `with_clause` group, so a `CREATE` written *before* a `WITH` cannot match. Real TCK queries that fail on exactly this:

```cypher
UNWIND [1,2,3,4,5] AS x CREATE (n:N {num: x}) WITH n WHERE n.num % 2 = 0 RETURN n.num
UNWIND [1,2,3,4,5] AS x CREATE ()-[r:R {num: x}]->() WITH sum(r.num) AS sum RETURN sum
CREATE (a) WITH a UNWIND [0] AS i CREATE (b) CREATE (a)<-[:T]-(b)
UNWIND ['a','b'] AS actor MERGE (m:Movie {name: actor}) …
```

Counted by leading construct: **66 involve `UNWIND … CREATE/MERGE … WITH`**, **41 involve `MERGE` in a position the rule does not allow**.

`match_stmt` has the same shape and the same limitation.

## Why this is not a small fix

Cypher's real grammar is a *sequence* of reading, writing and projecting clauses with very few ordering constraints — roughly `(reading | updating | WITH)* RETURN?`. Ours enumerates one permitted order per statement kind, which is why each new arrangement needs a new alternative and why the rules have grown to a single line of forty terms.

The AST matches the grammar's assumption: `create_clause` is `Option<CreateClause>`, `with_clause` is `Option<WithClause>` plus an `extra_with_stages` vector bolted alongside, and `with_split_index` records the one boundary the planner knows about. A query with `CREATE … WITH … CREATE` has two creates on opposite sides of a barrier and nowhere to put the second.

So the change is: model the query as an ordered `Vec<Clause>`, and have the planner walk it. That is a real refactor of the parser, the AST and the planner's entry point, and it should be done deliberately rather than by adding a fortieth alternative to `unwind_stmt`.

## Why it is worth doing

[...]

PR #619, opened 2026-08-18
SET n = {…} and SET n += {…} — whole-entity assignment

sandeepkunkunuru:
`SET n += $props` is the standard upsert idiom and neither spelling parsed. The right-hand side may be a map or another entity, so `SET a = b` copies b's properties onto a.

```cypher
MATCH (n:P) SET n += {b: 20, c: 3}   -- merge: keeps a, overwrites b, adds c
MATCH (n:P) SET n  = {z: 9}          -- replace: a and b are removed
MATCH (a:B), (b:A) SET a = b         -- copy b's properties, drop a's own
```

## What the tests are actually checking

The two operators differ **only** in what happens to properties the right-hand side does not mention. So every test asserts an absent key as well as the present ones — asserting only that the new values arrived would pass against an implementation that merged in both cases, which is the obvious way to get this half-right.

Two more tests pin the grammar's neighbours rather than the feature itself. `set_item` needs a `.` so `n = {…}` falls through to the entity form, and `set_entity_item` must precede `set_label_item` because both begin with a bare variable — so `SET n.a = 5` and `SET n:Admin` each get a test. A new grammar alternative is exactly the kind of change that silently reroutes the case sitting next to it.

Assigning a scalar to an entity is a type error rather than a silent no-op: it has no sensible meaning, and picking one would hide the mistake.

## Numbers

- openCypher TCK: **663 → 671** of 1,239 evaluated (54.2%), from 46.6% at the start of this sweep.
- `cargo test --workspace`: green across 93 blocks. 8 new tests.

Follow-on from #615. The remaining large TCK cluster is #617 (clause ordering), which is a parser/AST/planner refactor rather than another grammar alternative.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,3 +76,25 @@ jobs:
         # The sweep runs these; here we only prove they still build, which is
         # what actually breaks when a signature changes.
         run: cargo check --workspace --all-targets
+
+      # ---- Conformance gates (spec 18 §5: these block commits, not releases,
+      # because they are fast and catch the most damaging classes).
+      #
+      # Each is an ordinary example that exits non-zero on failure, so it
+      # gates without any bespoke CI machinery. `--release` because both are
+      # seconds either way and the debug build of the engine is slow enough
+      # that the dev profile dominates.
+
+      - name: CH-IMPORT — the same graph by different paths is the same graph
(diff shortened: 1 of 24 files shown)


sandeepkunkunuru:
Superseded by a branch rebuilt on top of main — #615 was squash-merged, so this branch's history diverged and the merge commit could not be created cleanly. Same diff, reopened as a new PR.

Issue #616, opened 2026-08-18
SF10: IC6 times out and IC11 is 11.5× Neo4j — the last two complex reads outside PERF-01


sandeepkunkunuru:
From the first same-host SF10 comparison (AWS r6a.8xlarge, 32 vCPU/246 GB, `facb1fa` vs Neo4j 5-community, identical `sf10.json` params, 29,987,835 nodes verified on both). Full writeup: `benchmarks/ldbc-snb-interactive/SF10-2026-08-19-SAME-HOST.md` in samyama-graph-competitor-benchmarks.

**Twelve of fourteen complex reads are now within 5× of Neo4j and six are faster.** These are the two that are not.

## IC6 Tag Co-occurrence — times out

```
IC6   Tag Co-occurrence   -   -   -   ERROR
       Query error: Runtime error: Query timed out after 10 rows
```

Neo4j answers it in **422 ms**. It is not that we are slightly slow — we do not finish.

Worth noting it *did* complete in July at 31.5 s on the older host, so this is the per-query timeout becoming the binding constraint, not a regression in the query. Whatever changed made the other deep reads 6–20× faster and left IC6 where it was.

IC6 is the tag co-occurrence query — friends-of-friends, then their posts' tags, excluding a given tag. The shape is a two-hop expansion feeding a grouped count, which is the same family as IC5 and IC9 (both now *faster* than Neo4j). That IC6 alone did not benefit is the interesting part and probably the fastest route to the cause.

## IC11 Job Referral — 11.5× Neo4j

| | Samyama | Neo4j |
|---|---:|---:|
| IC11 | 290 ms | 25.2 ms |

The only complex read that completes and is outside the 5× target. IC11 is friends-and-friends-of-friends filtered by a company's country and a work-from year — a two-hop expansion with a selective predicate on the far side. A candidate is that the predicate is not reaching the expansion, so the join is materialising far more rows than survive; `EXPLAIN` on the SF10 shape should say quickly.

## Why these two block the gate

`PERF-01`'s H1 target is **≤5× the best competitor on all** SNB-I complex reads. With these two outstanding it fails on 2 of 14 against Neo4j alone — and the target is against the best competitor, which the July run suggests is TigerGraph on several queries.

Reproducer:

```
python3 harness/run.py suite CH-BENCH-LDBC --ldbc-data <sf10> --ldbc-params config/ldbc-snb-interactive/sf10.json
```

sandeepkunkunuru:
[...]

Issue #618, opened 2026-08-18
PERF-04: short reads are no longer the fastest — IS7 is 41× FalkorDB on the same host


sandeepkunkunuru:
`PERF-04` says point and selective reads **remain the fastest of all compared engines**. On the first same-host SF10 run that includes an in-process competitor, they are not.

AWS r6a.8xlarge, SF10, same day, same data, identical `sf10.json` params. Samyama `facb1fa` embedded, Neo4j 5-community over Bolt, FalkorDB in-process.

| Query | Samyama | Neo4j | FalkorDB | best | ratio |
|---|---:|---:|---:|---:|---:|
| IS1 Person Profile | 0.08 | 9.86 | 0.08 | 0.08 | 1.03× |
| IS2 | 0.19 | 11.76 | 0.93 | 0.93 | 0.20× |
| IS3 Friends of Person | 1.50 | 12.03 | 0.53 | 0.53 | **2.81×** |
| IS4 | 0.04 | 9.04 | 0.08 | 0.08 | 0.51× |
| IS5 | 0.06 | 7.88 | 0.23 | 0.23 | 0.26× |
| IS6 | 0.09 | 7.93 | 0.34 | 0.34 | 0.26× |
| IS7 Replies to Post | **27.10** | 8.52 | 0.66 | 0.66 | **40.9×** |

## Why this was not visible before

The July SF10 comparison concluded "Samyama wins short/point reads (IS1–IS6 sub-millisecond, in-process embedded) — faster than every other engine on point lookups and 1-hop reads."

That comparison was mostly against engines paying an HTTP or Bolt round trip. Against Neo4j alone every short read still looks like a rout, and most of that margin is transport: a ~7 ms floor runs through every Neo4j number. **FalkorDB is also in-process**, and once transport is out of the way it is faster than us on IS3 and **41× faster on IS7**.

So the claim was not wrong about the measurements; it was wrong about what they meant.

## IS7 specifically

IS7 is "replies to a post": one hop to the replying comments, plus each reply's creator and whether that creator is a friend of the original author. 5 rows returned, 27 ms taken. Every other short read here is under 2 ms, so this is not a general short-read problem — it is one query shape.

FalkorDB at 0.66 ms and Neo4j at 8.52 ms *including* a Bolt round trip both suggest this should be well under a millisecond of actual work.

## Definition of done

[...]

PR #621, opened 2026-08-19
Pin a variable-length expand's target when it resolves to one node (IC6 timeout)

sandeepkunkunuru:
LDBC **IC6 times out at SF10** (#616). Neo4j answers it in 422 ms.

The plan is already right: it anchors on the selective tag, walks to the posts carrying it, then to their creators. What it does next is the problem — it asks, of each of thousands of candidate people, *"is this one specific person within two hops of you"*, and answers by expanding that candidate's entire two-hop neighbourhood and looking. Thousands of BFS runs to test membership of a single node.

One reversed BFS from the pinned node answers it for every row.

**IC6 at SF1: 110 ms → 3.7 ms.** (Host calibration 48 ms and 40 ms respectively, so at most 20% of that is the machine, not the change.) SF10 verification is running on a fresh spot instance; this PR will be updated with the result before merge.

## The guards, and why each is load-bearing

The planner sets the pin only when it can resolve the far endpoint to **exactly one node** by asking the property index — not by estimating. A predicate that merely narrows the variable is not a pin, and treating it as one drops rows silently.

- **`min_hops <= 1`** — a set keyed on shortest distance cannot answer `*2..n`, because a node whose shortest distance is 1 may still have a conforming longer walk.
- **no path variable** — a membership test produces no path to bind.
- **exactly one match** — two nodes sharing an indexed value must both stay reachable.

`EXPLAIN` now prints `[target pinned to node N]`. Whether this fires changes the operator's cost by orders of magnitude, and an optimisation you cannot see in a plan is one nobody can tell has stopped firing.

## Tooling that came out of the investigation

- **`--explain` on the LDBC bench** — prints plans *without running the queries*. `--profile` executes, which is useless for exactly the queries most worth inspecting: IC6 times out, so the one plan you most want is the one PROFILE cannot produce.
- **`SAMYAMA_EXPLAIN_ANCHORS=1`** — prints the cost of every candidate anchor. `EXPLAIN` shows only the winner, so when the winner looks wrong there is nothing else to inspect but the source.

That tracing earned itself immediately, in an instructive way: it showed the anchor for `p` costed at **zero rows**, which looked like a serious cost-model bug. It was not — the bench's default parameters name a person who does not exist in this extract, so the anchor really did match nothing, and the bench's own "0 empty" check had been saying so. I nearly fixed a non-bug. The real diagnosis came from re-running with `--derive-params`.

## Tests

The planner only takes this path when its cost model prefers an anchor other than the target, which depends on real statistics — on a fixture small enough to reason about, anchoring at the target is genuinely the better plan, and shaping the fixture to force otherwise would be dishonest. So the mechanism is tested **directly**: the same operator over the same graph, with and without the pin, must produce the same rows.

Reversing the BFS direction incorrectly — the obvious mistake — fails two of those tests; checked by making the change and running them.

`cargo test --workspace` green across 94 blocks. SF1: 21/21, 0 empty.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -28,6 +28,8 @@ cargo bench --bench mvcc_benchmark             # MVCC & arena allocation
 cargo bench --bench late_materialization_bench  # Late materialization traversal
 cargo bench --bench graph_optimization_benchmark # Metaheuristic optimization solvers
 cargo bench --bench ldbc_benchmark             # LDBC SNB Interactive queries (needs data)
+cargo bench --bench ldbc_benchmark -- --explain          # print plans, do NOT run the queries
+cargo bench --bench ldbc_benchmark -- --derive-params 50 # parameters valid for THIS extract
 cargo bench --bench ldbc_bi_benchmark          # LDBC SNB BI queries (needs data)
 cargo bench --bench finbench_benchmark         # LDBC FinBench queries (synthetic data)
 cargo bench --bench hierarchy_benchmark        # OEH index: build, order test, roll-up vs subtree size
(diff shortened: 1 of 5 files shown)


sandeepkunkunuru:
**Verified at SF10.** Re-ran on a fresh `r6a.8xlarge` at `6551dd1` with this branch. Host calibration **37 ms in both runs**, so this is a clean before/after.

| | before | after | Neo4j | FalkorDB | ratio to best |
|---|---:|---:|---:|---:|---:|
| **IC6 Tag Co-occurrence** | **timeout** | **219 ms** | 422 | 10 711 | **0.52× — fastest of the three** |

- **21 of 21 queries complete** (was 20 — IC6 errored).
- **13 of 14 complex reads within 5×** of the best same-host competitor (was 12).
- IC11 at 11.4× is now the only complex read outside the target.

Full writeup with provenance: `benchmarks/ldbc-snb-interactive/SF10-2026-08-19-SAME-HOST.md` in samyama-graph-competitor-benchmarks.

## Recorded rather than smoothed

The three deep reads moved 4–14% between the two runs at identical calibration: IC9 4 081 → 4 665 ms, IC5 5 132 → 5 350, IC1 680 → 699. That is inside run-to-run variance for queries of this length, but the second run also carries the anchor-model change in this branch, so it is stated rather than waved off. If it is a real regression it will show up on the next SF10 pass; `CH-REGRESS` exists now to catch that class on SF1.

## What this does not do

It does not meet PERF-01. IC11 is still 11.4×, IS7 is 40× on the short-read side (#618), and **TigerGraph has not been re-run on this host** — in July it was fastest on exactly IC1 and IC9.

PR #622, opened 2026-08-19
Clauses in the order Cypher allows: clause-sequence parser and planner (#617)

sandeepkunkunuru:
Every statement rule in the grammar encodes one permitted clause order, with writes at the end. Cypher does not work that way, so `MATCH (n) SET n.x = 1 WITH n RETURN n.x` was a syntax error.

**TCK 671 → 706** of 1,239 evaluated (54.2% → 57.0%), errored 242 → 202.

## The scope in #617 was wrong, and I corrected it before starting

The issue claimed 107 scenarios. Bisecting each of the 162 unparseable queries to its first failing clause, then re-testing that clause **in a legal position**, separates two different problems:

| | count |
|---|---:|
| **Ordering** — the clause is fine elsewhere | **47** |
| **Content** — the clause fails wherever you put it | 115 |

The fallback rescues **69** in total, because some "content" failures were really "no statement rule can *start* with this clause". The remaining 115 are a separate backlog: pattern comprehensions, temporal constructors, map projections, aggregates in compound expressions.

## A second entry point, not another alternative

PEG ordered choice commits to the first alternative that matches a **prefix**. `create_stmt` accepts `CREATE (a) WITH a …` up to the `WITH`, at which point the choice is committed and no later alternative is tried — so adding a general rule to `statement` did literally nothing. `parse_query` now tries `query` first and falls back to `pipeline_query` only on failure. Every query the established rules accept keeps taking them and keeps producing the by-kind AST the planner has always used; a test asserts that seven common query shapes never reach the fallback.

## The syntax was not the real blocker

Making the grammar accept these queries got as far as parsing, planning correctly, running — **and not writing**.

The default `next_mut` on a pass-through operator delegates to `next`, which reads its input read-only. A materialising operator therefore severed mutability for everything below it. `MATCH (n) SET n.x = 1 WITH n RETURN n.x` returned the *old* value, the store was unchanged, and nothing reported a problem.

That is the actual reason the grammar only ever allowed writes after the last projection — not syntax, but that mutability never reached below a barrier.

`WithBarrierOperator` and `AggregateOperator` now drain their input mutably into a `MaterializedOperator` before aggregating. Both materialise their whole input anyway, so nothing is buffered that would not have been, and their grouping code — the most intricate in the file — is untouched.

## Unsupported orders are refused, not mis-planned

CREATE and MERGE inside the pipeline, and a MATCH after a WITH (7 of the 69), are not threaded yet. The by-kind fields are empty for these queries *by construction*, so falling through to the established path would read them as "no CREATE at all" and answer a different question. The error names the offending clause and the query's shape.

## Tests

They assert **the graph after**, not the rows returned. A write that reports success and does nothing is invisible to any test that only reads the result set, and that was the bug. Removing the barrier's mutable pull fails four of the nine; checked by making the change and running them.

`cargo test --workspace` green across 94 blocks; both CI conformance gates pass.

Also adds `examples/parse_check.rs` — reads queries from stdin, prints `ok`/`fail` per line. The scope analysis needed to parse a few thousand variants and a process per query makes that unbearable.

### DIFF
--- a/examples/parse_check.rs
+++ b/examples/parse_check.rs
@@ -1,0 +1,29 @@
+//! Parse queries from stdin, one per line; print `ok` or `fail` per line.
+//!
+//! A pipe rather than a flag because the useful question is rarely "does this
+//! one query parse" — it is "which of these two hundred variants parse", asked
+//! by a script that is bisecting a construct. Starting a process per query
+//! makes that unbearably slow.
+//!
+//!   printf '%s\n' 'MATCH (n) RETURN n' 'CREATE ()' | cargo run --release --example parse_check
+
+use samyama::query::parser::parse_query;
+use std::io::{self, BufRead, Write};
+
+fn main() {
+    let stdin = io::stdin();
+    let stdout = io::stdout();
+    let mut out = stdout.lock();
+    for line in stdin.lock().lines() {
+        let Ok(line) = line else { break };
+        if line.trim().is_empty() {
+            let _ = writeln!(out, "skip");
+            continue;
+        }
+        let verdict = match parse_query(&line) {
+            Ok(_) => "ok",
+            Err(_) => "fail",
+        };
+        let _ = writeln!(out, "{verdict}");
+    }
+}
--- a/src/query/ast.rs
+++ b/src/query/ast.rs
@@ -121,6 +121,23 @@ pub struct Query {
     pub foreach_clause: Option<ForeachClause>,
     /// UNWIND clause (optional)
     pub unwind_clause: Option<UnwindClause>,
+    /// Every clause of the query, in the order it was written.
+    ///
+    /// The fields above describe a query by *kind* — all the MATCHes here, the
+    /// CREATE there — which is enough only while the grammar fixes one legal
+    /// order. Cypher does not: a write may sit before a `WITH`, and two writes
+    /// may be separated by a projection, and neither has anywhere to live in a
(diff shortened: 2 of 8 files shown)


sandeepkunkunuru:
**CREATE now threaded through the pipeline too.** TCK **711** of 1,244 evaluated (57.2%), up from 671/54.2% at the start of this branch. Coverage 76.7% → 77.0%, since five fixtures that could not be built before now parse and their scenarios are judged rather than skipped.

`build_create_on_input` is extracted from the `MATCH … CREATE` path so both build the same operator — the pipeline does not grow a second implementation of node-and-edge creation to drift from it.

Two ordering details, both of which produce a **silently wrong graph** rather than an error when got backwards:

- **Scope is read before the CREATE, not after.** A CREATE references variables already in scope and creates the rest. Adding the pattern's own variables first makes them all look already-bound and the clause creates *nothing* — query succeeds, graph unchanged. `CREATE (a:A) WITH a CREATE (a)-[:R]->(:B)` distinguishes the two and has a test.
- **The "data write with no RETURN returns no rows" rule reads the by-kind fields**, which are empty for a pipeline query. Without including `clauses`, `CREATE (a) WITH a CREATE (b)` returned a row where `CREATE (a), (b)` correctly returns none.

Still refused by name: MERGE in the pipeline, and a MATCH after a WITH (7 scenarios). Those are the next increment.

`cargo test --workspace` green across 95 blocks; both CI gates pass locally.

Issue #689, opened 2026-08-21
Temporal types collapse to a millisecond DateTime — no Date/Time/LocalTime/LocalDateTime, no nanoseconds


sandeepkunkunuru:
Every temporal value in the engine is `PropertyValue::DateTime(i64)`, a Unix timestamp in **milliseconds**. Cypher has five distinct temporal types and nanosecond precision, so `date()`, `time()`, `localtime()`, `localdatetime()` and `datetime()` are all indistinguishable once evaluated.

## Evidence

`localtime({hour: 10, minute: 35})` and `localtime({hour: 12, minute: 31, second: 14, nanosecond: 645876123})` both become `DateTime(millis)`. The TCK expects them rendered as `'10:35'` and `'12:31:14.645876123'`.

Two separate problems, and the second is the harder one:

1. **Type identity.** A `Date` is not a `DateTime` is not a `LocalTime`. They render differently, compare within their own kind, and `toString` differs.
2. **Precision.** `645876123` nanoseconds cannot be represented in an `i64` of milliseconds at all. This is not a formatting issue — the value is destroyed at construction.

## Cost

10 openCypher TCK scenarios — WithOrderBy1 [11]–[20], the "sort dates / local times / times / local date times / date times" family. Ordering already works; they fail on the rendered value.

Diagnosed while working the ORDER BY cluster: those ten looked like sort bugs and are not. `UNWIND [date(1985...), date(1910...)] ... ORDER BY d LIMIT 1` already returns the 1910 value. The sort is right; the type is wrong.

## Why this is not a small change

`PropertyValue` is the engine's value type. Adding five variants touches:

* the enum and every exhaustive `match` on it — which is the point of the design, per the module comment;
* `Ord` and `cypher_order` (added in #688), which need per-type comparison;
* serde, and therefore **the snapshot format** — an on-disk compatibility question, not just a code one;
* the columnar store's per-type property arrays;
* every temporal function, plus `toString` and result rendering.

That is why this is filed rather than attempted: it is a value-type and storage-format change, and starting it halfway would leave the snapshot format in an ambiguous state.

## Suggested sequence

1. Decide the representation first — in particular whether `DateTime` keeps `i64` millis for compatibility and the new types carry their own precision, or whether everything moves to (seconds, nanos). This is the on-disk decision and everything else follows it.
2. Add the variants with `Ord` and `cypher_order` arms, behind exhaustive matches so nothing is silently missed.
[...]

Issue #710, opened 2026-08-23
A var-length segment returns shortest paths, not paths — wrong answers in both directions


sandeepkunkunuru:
`VarLengthExpandOperator::expand_from` is a breadth-first search with a **node**-visited
set:

```rust
self.for_each_neighbor(cur, type_filter, store, |nb, eid| {
    if visited.insert(nb) {          // <- a node is discovered once, at its shortest depth
        parent.insert(nb, (cur, eid));
        next.push(nb);
    }
});
```

openCypher specifies something different. `-[:R*m..n]-` matches every path whose
**relationships** are distinct; nodes may repeat. So the operator is answering
"which nodes are reachable in m..n hops" when the question is "what paths of
length m..n exist", and those differ in both directions.

## Measured

`tests/var_length_path_enumeration.rs` (in this commit, `#[ignore]`d against this
issue). Graph is a triangle — `a-b`, `b-c`, `a-c`, all `:R`:

| query | openCypher / Neo4j 5 | Samyama | |
|---|---|---|---|
| `MATCH (a {name:"a"})-[:R*1..2]-(x) RETURN x.name` | `b, b, c, c` | `b, c` | **under-reports** |
| `MATCH (a {name:"a"})-[:R*2..2]-(x) RETURN x.name` | `b, c` | *nothing* | **returns an empty result** |
| `MATCH (a {name:"a"})-[:R]-(y)-[:R*1..1]-(z) RETURN z.name` (one edge only) | *nothing* | `a` | **over-reports** |

The second row is the one to look at. `*2..2` from `a` matches `a-b-c` and
`a-c-b`; both endpoints sit at distance 1, the BFS marks them visited at depth 1,
and depth 2 never reconsiders them. The query returns **zero rows and no error**.

The third row is #684's rule at a boundary it did not cover: `ExpandOperator`
now tracks used edges per clause, but `VarLengthExpandOperator` neither reads
`record.used_edge_slice()` nor calls `mark_edge_used`, so edges leak across the
operator boundary in both directions.

## The TCK already says so

`clauses/match/Match6.feature` scenario **[14] Named path with undirected fixed
variable length pattern**:

```cypher
MATCH topRoute = (:Start)<-[:CONNECTED_TO]-()-[:CONNECTED_TO*3..3]-(:End)
RETURN topRoute
```

over a graph where `mid` has two parallel edges to `:End` and two to `other`.
Expected: **four rows**, every one of them revisiting `mid` and ending on an
[...]

PR #757, opened 2026-08-24
Expand Scenario Outlines: TCK coverage 77.0% → 96.5%, and the pass rate was measured on the easy half (#756)

### LINKED ISSUE #756: The TCK harness skips every Scenario Outline: 274 of 1,615 scenarios, and the 87.1% pass rate is measured without them
## What

`examples/tck_runner.rs` reports `Scenario Outline` as an unsupported step and skips it:

```rust
if line.starts_with("Scenario Outline:") {
    // Outlines need Examples expansion, which this harness does not
    // do. Reported rather than silently dropped.
    s.unsupported = Some("Scenario Outline".into());
}
```

That is **274 scenarios — the largest single skip reason, larger than every other reason combined** (39 user-defined procedure, 34 parameters, 19 named fixture graph, 5 setup did not run).

An outline is a template: its steps carry `<placeholder>` tokens and each row of its `Examples:` table is one concrete scenario. So 274 skipped outlines are not 274 unjudged scenarios — they are about **2,280**.

## Why it matters

The published figure is `CH-TCK 87.1% of 1,244 evaluated (77.0% coverage)`. Expanding the outlines and changing nothing else:

| | before | after |
|---|---|---|
| scenarios | 1,615 | 3,897 |
| evaluated | 1,244 (77.0%) | 3,761 (96.5%) |
| pass | 1,083 | 2,103 |
| **pass rate** | **87.1%** | **55.9%** |

Nothing regressed — the pass *count* nearly doubled. The 87.1% was measured on a sample that excluded the outlines, and the outlines are the harder cases: they are how the TCK enumerates the edge cases of a feature it has already established the happy path for.

The harness's own comment on the ratchet predicted this exactly:

> a scenario the harness learns to judge shifts a scenario from "skipped" into "evaluated", which changes the percentage without the engine changing at all

This is that event. It is a **measurement correction, not a regression**, and the H1 `CH-TCK >= 85%` gate does not hold against the corrected measurement.

## Where the 1,658 failures actually are

```
    322 of 322   (19.4% of all failures)  Temporal9
    207 of 207   (12.5%)                  Temporal1
    183 of 183   (11.0%)                  Temporal3
    131 of 131   ( 7.9%)                  Temporal10
     82 of 97    ( 4.9%)                  Match6
     77 of 86    ( 4.6%)                  Match1
     74 of 86    ( 4.5%)                  Match2
     59 of 96    ( 3.6%)                  WithOrderBy1
     53 of 53    ( 3.2%)                  Temporal2
     46 of 93    ( 2.8%)                  WithOrderBy3
   1658 across 96 features in total
```

`Temporal*` is **980 of the 1,658 — 59% of every TCK failure — and it is one open issue, #689.** Excluding it the engine is at 2,097/2,775 = **75.6%**. That single number reorders the correctness backlog: #689 is not one feature gap among many, it is the majority of the remaining conformance work.

The rest are real and specific, e.g.:

* `Match1 [10] Fail when a path has the same variable in the same pattern` — expects `SyntaxError`, we return 0 rows. Missing *validation*, not missing execution.
* `ExistentialSubquery1..3` — `WHERE exists { ... }` does not parse.
* `Match3 [8] MATCH (a)-[:T|:T]->(b)` — a repeated type in an alternation does not parse.
* `Map1 [5] map.\`backtick quoted\`` — delimited identifiers after a dot do not parse.

## Found by

Asking which check nobody runs. The harness reports its own gaps honestly — the skip histogram was right there and said `274 Scenario Outline` at the top of the list every run. It was read as "274 scenarios" rather than "the harness cannot see the majority of the corpus".

sandeepkunkunuru:
Closes #756.

## The change

`examples/tck_runner.rs` skipped every `Scenario Outline:`. An outline is a template — its steps carry `<placeholder>` tokens and each `Examples:` row is one concrete scenario — so the 274 skipped outlines were about **2,280 cases the engine had never been judged on**. This expands them.

## What it does to the numbers

| | before | after |
|---|---|---|
| scenarios | 1,615 | 3,897 |
| evaluated | 1,244 (77.0%) | 3,761 (**96.5%**) |
| pass | 1,082 | **2,102** |
| pass rate | 87.1% | **55.9%** |

**Nothing regressed — the pass count nearly doubled.** The 87.1% was measured over a sample that left the outlines out, and the outlines are the harder half: they are how the TCK enumerates the edge cases of a feature whose happy path it has already established.

The ratchet's own comment predicted this event verbatim:

> a scenario the harness learns to judge shifts a scenario from "skipped" into "evaluated", which changes the percentage without the engine changing at all

So the CI floor moves `1,082 → 2,102` as a **re-baseline, not a gain**. Both the runner and the workflow now spell out which of the two a floor move is, because a floor that jumps by a thousand reads as progress unless the commit says otherwise.

## The H1 gate

`CH-TCK >= 85%` does not hold against the corrected measurement, and this PR does not pretend it does — the runner prints `NOT MET` and, underneath it, the before/after pair so the two figures are never quoted apart. Scoping the gate back to non-outline scenarios to keep it green was the alternative and would have been the wrong one.

## Where the failures actually are

A bare 55.9% is not actionable, so the report gained a failure-location table:

```
    322 of 322   (19.4% of all failures)  Temporal9
    207 of 207   (12.5%)                  Temporal1
    183 of 183   (11.0%)                  Temporal3
    131 of 131   ( 7.9%)                  Temporal10
     82 of 97    ( 4.9%)                  Match6
     77 of 86    ( 4.6%)                  Match1
     74 of 86    ( 4.5%)                  Match2
     59 of 96    ( 3.6%)                  WithOrderBy1
     53 of 53    ( 3.2%)                  Temporal2
     46 of 93    ( 2.8%)                  WithOrderBy3
   1658 across 96 features in total
```

`Temporal*` is **980 of 1,658 — 59% of every TCK failure — and all of it is one open issue, #689.** Excluding it the engine sits at 2,097/2,775 = **75.6%**. That single line reorders the correctness backlog.

## A bug this change shipped and then fixed

The first version cleared `unsupported` on every expanded clone. Thirteen `Call5` scenarios stopped being *"skipped: user-defined procedure"* and became *"errored: Unknown procedure"* — the expansion converting a reported **harness** gap into a reported **engine** defect. Outline-ness is now a separate `is_outline` flag instead of a preset `unsupported` marker, so every step handler behaves exactly as it does for an ordinary scenario.

## Tests

Five new, all in `examples/tck_runner.rs`:

- `an_outline_expands_to_one_scenario_per_example_row` — including substitution into the *expected* table, which a harness that interpolated only the query would get wrong in the direction of blaming the engine
- `example_rows_do_not_land_in_the_expected_table`
- `several_examples_blocks_each_expand_against_their_own_header`
- `an_outline_without_examples_is_skipped_not_dropped` — dropping it would shrink the denominator and inflate coverage for work not done
- `expansion_does_not_erase_a_reason_the_scenario_cannot_run` — the `Call5` regression above

The first two were checked against a deliberately broken parser before being trusted. Removing the `continue` from the `Examples:` table arm did **not** fail them — the match falls through to `_ => {}` — so the assertion was moved onto the line that is actually load-bearing. Breaking that line fails three of the five.

`cargo test --workspace --release`: green. `cargo test --release --example tck_runner`: 9 passed.

## Downstream

`--emit-scenarios` and `--judge` now emit the expanded set, so the competitor SCORECARD will compare ~3,761 scenarios instead of ~1,244. That is the point of the change, but it means the benchmarks repo's recorded numbers need a re-run against this commit rather than a merge of old and new.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,7 +134,7 @@ jobs:
         # why in the same commit.
         run: |
           git clone --depth 1 --branch 2024.3             https://github.com/opencypher/openCypher.git /tmp/openCypher
-          cargo run --release --example tck_runner --             --features /tmp/openCypher/tck/features --min-pass 1083
+          cargo run --release --example tck_runner --             --features /tmp/openCypher/tck/features --min-pass 2103
 
       - name: CH-DETERM-THREADS — the answer does not depend on the thread count
         # LANG-14 (#611). CH-DETERM varies the *process*, which catches a
--- a/examples/tck_runner.rs
+++ b/examples/tck_runner.rs
@@ -4,6 +4,12 @@
 //! had never been run — a "~90% OpenCypher coverage" claim was withdrawn as
 //! unmeasured in #437. This produces the measurement.
 //!
+//! `Scenario Outline:` blocks are expanded against their `Examples:` rows
+//! (#756). They were skipped until then, which was 274 of the corpus's 1,615
+//! scenarios — more than every other skip reason combined, and ~2,280
+//! concrete cases once expanded. The pass rate quoted before that expansion,
+//! 87.1%, was measured over the 1,244 scenarios that remained.
+//!
 //! ## What it does and does not attempt
 //!
 //! The TCK is Gherkin. This implements the step vocabulary that covers the
@@ -20,7 +26,7 @@
 //!
 //!   * **pass rate over evaluated scenarios** — what the engine gets right
 //!     among the scenarios this harness can actually judge;
-//!   * **coverage** — how many of the 1,615 scenarios it can judge at all.
+//!   * **coverage** — how many of the corpus's scenarios it can judge at all.
 //!
 //! Both belong in any quote of the result.
 //!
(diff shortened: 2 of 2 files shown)


sandeepkunkunuru:
Rebased onto `main` now that #760 (#710) has landed.

The floor moves with it: #710 earned one scenario, so the re-baseline is **1,083 → 2,103**, not 1,082 → 2,102. Re-measured on the rebased branch rather than adjusted arithmetically:

```
  evaluated          3761  (96.5% of scenarios)
    pass             2103
  PASS RATE          55.9%  of evaluated scenarios
```

CI floor and the runner's re-baseline comment both updated to 2,103. Workspace suite re-running on the rebased tree — the pre-rebase run was exit 0 with 3,297 passing, but a rebase is a different tree and the earlier number does not carry over.

sandeepkunkunuru:
Rebased onto `main` after #762 (#761) landed. That conflicted — both branches add tests at the top of `mod tests` — and the resolution keeps **both** sets; all 11 pass together.

Re-measured on the rebased branch rather than carrying the old number forward:

```
  evaluated          3761  (96.5% of scenarios)
    pass             2103
  PASS RATE          55.9%  of evaluated scenarios
```

CI floor stays 2,103. `cargo test --workspace --release`: exit 0, **3,300 passed, 0 failed**.

Worth noting the two land well together: #761 fixed a parser that hung on unrecognised engine output, and this PR triples the corpus the parser is fed. Expanding outlines without that fix would have made the hang much easier to hit.

Issue #798, opened 2026-08-25
A pattern predicate in WHERE may introduce variables that nothing binds


sandeepkunkunuru:
## What

```cypher
MATCH (n) WHERE (n)-[r]->(a) RETURN n     -- expected SyntaxError: UndefinedVariable
```

`r` and `a` are bound nowhere. A pattern used as a predicate is a **test** — it asks whether such an edge exists — and the names inside it have no meaning outside. `EXISTS { ... }` is the form that may introduce names.

15 scenarios in `Pattern1` [10]. We answered all of them.

## A trap this exposed

A bare pattern predicate and a real `EXISTS { ... }` **desugar to the same AST node**. The parser says so:

> `WHERE (:Acc)-[:SUPPORTS]->(o)` means "such a path exists", which is exactly `EXISTS { MATCH ... }` — so it desugars to the same node and inherits its (already correct) evaluation

That is right for *evaluation* and wrong for *validation*: one may introduce variables and the other may not. My first version enforced the rule on both and **rejected every `EXISTS` in the codebase**. `ExistsSubquery` now carries `bare_pattern`, and adding the field made the compiler surface all ten match sites.

## My own test should have caught that and did not

`exists_subqueries_may_introduce_variables` was written as *"if it fails, check the reason"*:

```rust
if let Err(e) = parse_query(q) { assert!(!msg.contains("used as a predicate")); }
```

**A test that tolerates failure cannot report one.** It passed while the rule broke every `EXISTS`; the workspace suite caught it instead, via a pre-existing parser test. It now asserts acceptance outright.

## Two scope limits, stated rather than papered over

* `WHERE (a)` parses as a **parenthesised variable**, not a pattern predicate — the grammar's `pattern_predicate` needs at least one edge. Refusing it is a different rule (undefined variable in `WHERE`), so **one TCK row still fails** and this issue stays open for it. Widening the check to claim 15/15 would be tuning the rule to the score.
* `EXISTS { MATCH ... RETURN ... }` does not parse at all — a pre-existing gap the TCK tracks as `ExistentialSubquery2`, unrelated to this.

## Measured

**3,273 → 3,287 (+14), zero regressions.**

sandeepkunkunuru:
Verified fixed on main (`b789e37`), probed with a two-node graph:

| query | result |
|---|---|
[...]

Issue #785, opened 2026-08-25
UNWIND … WITH … UNWIND loses the second UNWIND: VariableNotFound on a variable the WITH just defined


sandeepkunkunuru:
## Minimal repro

```cypher
UNWIND [1,2] AS a WITH [3,4] AS b UNWIND b AS c RETURN c
```

```
RuntimeError: VariableNotFound("b")
```

`b` is defined by the `WITH` immediately before the `UNWIND` that reads it.

## What works, and what does not

```cypher
WITH [3,4] AS b UNWIND b AS c RETURN c                      -- OK, 2 rows
MATCH (n) WITH [3,4] AS b UNWIND b AS c RETURN c            -- OK
UNWIND [1,2] AS a UNWIND [3,4] AS b RETURN b                -- OK, 4 rows
UNWIND [1,2] AS a WITH [3,4] AS b UNWIND b AS c RETURN c    -- ERR VariableNotFound("b")
UNWIND [1,2] AS a WITH a AS b UNWIND [5] AS c RETURN c      -- ERR VariableNotFound("c")
```

The last line is the clearest: the **unwind variable itself** is missing, so that `UNWIND` is not being planned at all. A leading `UNWIND` is what triggers it — with a leading `MATCH`, or with no leading clause, the same shape works.

## Cause

The parser loses the second `UNWIND`'s **position relative to the `WITH`**:

```
UNWIND [1,2] AS a WITH [3,4] AS b UNWIND b AS c RETURN c
  unwind_clause      = Some("a")
  extra_unwind       = ["c"]        <-- the post-WITH unwind, position discarded
  with_clause        = Some(1 item)
  extra_with_stages  = []           <-- empty
```

Both unwinds end up in flat fields with no ordering against the `WITH`, so the planner runs them before it, and `b` does not exist yet.

`extra_with_stages` already has the slot for this — its tuple is

```rust
Vec<(WithClause, Option<UnwindClause>, Vec<MatchClause>, Option<WhereClause>)>
```

and that `Option<UnwindClause>` is exactly the post-WITH unwind. Here it is never populated; the stage list is empty.

By contrast, `WITH [3,4] AS b UNWIND b AS c` parses into the **clause pipeline** (`clauses` populated, by-kind fields empty), where order is intrinsic — which is why it works.

## Suggested fix

Emit a `WITH` stage carrying its trailing `UNWIND` rather than flattening into `extra_unwind_clauses`. The grammar already models it:

```
[...]

Issue #769, opened 2026-08-24
Namespaced function calls do not parse: 453 TCK scenarios, and duration.between was implemented but unreachable


sandeepkunkunuru:
## What

`function_name` in `src/query/cypher.pest` has no dot:

```
function_name = @{ (ASCII_ALPHA | "_") ~ (ASCII_ALPHANUMERIC | "_")* }
```

So every namespaced function — `date.truncate`, `datetime.truncate`, `duration.between`, `duration.inSeconds` — is a **parse error**, not a missing implementation.

## Scale

**453 scenarios**, the entirety of two features:

* `Temporal9` — truncation, 322
* `Temporal10` — `duration.between` and friends, 131

Every one of them failed before reaching the evaluator.

## The part worth noticing

`duration.between` **is implemented**, and has been:

```rust
"duration_between" | "duration.between" => { ... }
```

That arm could never be reached, because nothing spelled with a dot could get past the parser. A function was written, tested through its underscore alias, and shipped unreachable through its actual Cypher spelling.

The same shape as #758, where the Rust side handled `ASCENDING` and the grammar never let one through. Two independent cases now of *the implementation being ahead of the grammar*, which suggests the gap is worth a sweep rather than another one-off: anything reachable only via a spelling the grammar cannot produce is dead code that looks live.

## Fix

Allow one namespace segment. Safe against property access because `function_call` requires a `(` immediately after the name, so `n.name` fails the rule and the parser backtracks as before — and `function_call` is already ordered ahead of `property_access` in `primary`.

Measured: **+19 from the grammar alone, no new code.** With truncation and duration rendering implemented on top, **2,570 → 2,889 (+319), zero regressions**.

## Found by

Reading the failure *detail* rather than the failure count. The summary said "322 Temporal9 failures"; the detail said `parse:` on every line. A feature that is 100% failing is more often one missing mechanism than 322 missing behaviours, and the histogram cannot tell you which.

sandeepkunkunuru:
Verified fixed on main (`b789e37`). `cypher.pest` now accepts a dotted `function_name` (comment at `:400`), and each of these parses and runs:

| query | result |
|---|---|
| `RETURN duration.between(date('2020-01-01'), date('2020-01-03'))` | `Duration { days: 2 }` |
| `RETURN date.truncate('month', date('2020-05-17'))` | `Date(18383)`, i.e. 2020-05-01 |
[...]

PR #822, opened 2026-08-25
fix(query): make CREATE/MERGE bound-variable checks WITH-aware (#764)

### LINKED ISSUE #764: CreateOnBoundVariable and MergeOnBoundVariable are not WITH-aware: a valid query is rejected
## What

```cypher
MATCH (a)-[r]->(b) WITH a CREATE (r:X) RETURN a
```

```
Semantic error: Variable `r` already declared; CREATE cannot add labels or properties to it
```

`r` is **not projected through the `WITH`**, so it is out of scope on the far side. The `CREATE` binds a *fresh* node that happens to reuse the name, which is legal Cypher. Same for `MERGE`:

```cypher
MATCH (a)-[r]->(b) WITH a MERGE (r:X) RETURN a
-> MERGE cannot impose new labels or properties on `r`, which is already bound; use SET instead
```

A fresh name in the same position is accepted, which confirms the rule is firing on the *name* rather than on anything about the clause:

```cypher
MATCH (a)-[r]->(b) WITH a CREATE (z:X) RETURN a   -- OK
```

## Why it matters more than it looks

`validate.rs` opens by saying scope analysis is deliberately absent because getting it slightly wrong

> would reject valid queries, which is a far worse failure than accepting an invalid one

This is that failure. It is not a missing check — it is a check that fires when it should not, and a user hitting it has no way forward except renaming a variable for no reason.

The rule itself is right; `MATCH (a) WITH a CREATE (a:Foo)` **is** an error, because `a` *is* carried through. What is missing is that a `WITH` re-scopes: a name it does not project is free again.

## Where

`CreateOnBoundVariable` / `CreateOnBoundRelationship` and `MergeOnBoundVariable` in `src/query/validate.rs`. They collect bound names without applying the `WITH` boundary, unlike `validate_variable_kinds` (#764), which does — `carry_kinds_through_with` keeps only names the `WITH` projects, and drops the rest.

## Suggested fix

Reuse that boundary. The kind checker already computes exactly the right set at each `WITH`; the bound-variable checks should consult the same scope rather than a flat accumulation of every name ever seen.

Two shapes to be careful of, both of which #764 had to handle:

* `WITH r AS q` renames — the binding survives under the new name.
* `WITH count(r) AS r` recomputes — the name survives but is no longer an entity, which is the case `VariableTypeConflict` already covers for collections.

## Found by

Writing `tests/variable_kind_conflict.rs` for #764. I asserted that the CREATE form is accepted, it failed, and the failure reproduced on `main` — so it is pre-existing and unrelated to that change. Rather than delete the assertion I inverted it into `create_and_merge_after_with_are_over_rejected_by_an_older_rule`, which records today's behaviour deliberately and **turns red the moment this issue is fixed**. Whoever fixes it should expect that test to fail and should flip it back.

56steve:
## Summary

Fixes #764: a valid query was rejected because `CreateOnBoundVariable` / `MergeOnBoundVariable` are not `WITH`-aware.

```cypher
MATCH (a)-[r]->(b) WITH a CREATE (r:X) RETURN a
-- before: Semantic error: Variable `r` already declared; CREATE cannot add labels or properties to it
-- after:  accepted — r is not projected through the WITH, so CREATE binds a fresh node
```

Same for the `MERGE` form. A fresh name in the same position (`… CREATE (z:X)`) was always accepted, which confirmed the rule was firing on the **name** rather than on scope.

As `validate.rs` states up front, over-rejecting a valid query is the worse failure — and a user hitting this had no way forward but to rename a variable for no reason.

## Root cause

`write_patterns` accumulated every name ever bound without applying the `WITH` boundary, unlike `validate_variable_kinds`, which re-scopes at each `WITH` through `carry_kinds_through_with`.

## Fix

- Add `carry_names_through_with` (the bound-name sibling of `carry_kinds_through_with`): only names a `WITH` projects *by name* stay bound; renames (`WITH a AS b`) carry under the new name.
- Apply it in **both** AST shapes inside `write_patterns` — the clause pipeline (`Clause::With`) and the by-kind fields (`with_clause`, split at `with_split_index`).
- `WITH *` is expanded to explicit items *before* validation runs (`expand_stars` precedes `validate` in the parser), so it carries names forward correctly and is not a loophole.

The rule is unchanged where it is right: `MATCH (a) WITH a CREATE (a:Foo)` is still rejected because `a` **is** carried through.

## Test plan

- [x] `tests/variable_kind_conflict.rs` — flips the pinned `create_and_merge_after_with_are_over_rejected_by_an_older_rule` (which existed to turn red the moment #764 was fixed) into `create_and_merge_after_with_may_reuse_a_dropped_name`, and adds `a_create_on_a_name_carried_through_with_is_still_rejected` covering the carried-through, rename, and `WITH *` guards. 9/9 pass.
- [x] `tests/clause_pipeline.rs` — 21/21 pass (carried-through re-mention cases like `CREATE (a) WITH a CREATE (b) …` unaffected).
- [x] `cargo test --workspace --no-fail-fast` — **3,448 passed, 0 failed**.

Closes #764.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,9 +132,16 @@ jobs:
         #
         # Raise the floor when it goes up; lowering it is allowed but must say
         # why in the same commit.
+        #
+        # Lowered 3665 -> 3656 on 2026-08-26 (then 3658 with #890): the harness began **checking
+        # declared side effects** (#888). Nothing regressed — 13 scenarios that
+        # had been passing on an empty result while the write they exist to test
+        # went unchecked now fail, including `MERGE (a)` creating a second node
+        # instead of matching the one that exists. Same shape as the outline
+        # expansion in #756: the ruler got honest and the count fell.
         run: |
           git clone --depth 1 --branch 2024.3             https://github.com/opencypher/openCypher.git /tmp/openCypher
-          cargo run --release --example tck_runner --             --features /tmp/openCypher/tck/features --min-pass 3428
+          cargo run --release --example tck_runner --             --features /tmp/openCypher/tck/features --min-pass 3700
 
       - name: CH-DETERM-THREADS — the answer does not depend on the thread count
         # LANG-14 (#611). CH-DETERM varies the *process*, which catches a
@@ -150,6 +157,13 @@ jobs:
         # after — a determinism gate added afterwards cannot tell a new bug
         # from a pre-existing one.
         run: ./scripts/check_thread_determinism.sh /tmp/openCypher/tck/features
+
+      - name: The release gate refuses for three reasons, not one
+        # It is the missing and the stale verdict that actually happen, and
+        # both look like "nothing is wrong" to a check that only compares a
+        # status string. This runs in CI because a gate nobody tests is a gate
+        # nobody can rely on at the moment it matters.
+        run: python3 scripts/test_check_release_gate.py
 
       - name: LINT-RATCHET — the clippy warning count may not rise
         # A ceiling, not a clean-tree gate: the debt stops growing while what
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,113 @@
+name: Prepare Release
+
+# Stage 1 of 2 in closing the "cut a release" loop. Until now, every release
+# has been a person hand-editing three version files across two languages,
+# running `git tag`, and filling in the GitHub Release UI — the actual source
+# of the staleness this repo's whole publish pipeline was built to fix. This
+# workflow does the bump and opens the PR; tag-release.yml (stage 2) does the
+# tag and the Release once that PR lands on main. Nothing publishes from this
+# workflow directly — main requires a reviewed PR and a green CI check, and
+# this respects that rather than bypassing it.
+#
+# Usage: Actions -> Prepare Release -> Run workflow -> version: 1.7.2
+
+on:
+  workflow_dispatch:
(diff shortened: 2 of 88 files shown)


sandeepkunkunuru:
Rebased onto `main` at `5c8b177`, which is eleven merges ahead of where this branch was cut.

Re-verified after the rebase:

- **TCK 3,561 → 3,561**, regression diff against the merge base **empty**. No change is the expected result: this fixes an over-rejection of a *valid* query, and the TCK does not exercise the shape. The measurement is here to show the change costs nothing, not to claim it gains anything.
- `variable_kind_conflict`, `property_access_targets`, `pattern_predicate_scope`, `boolean_operands`, `order_by_scope` — 33 tests, all pass.

Closes the third of the four issues deliberately left open at the end of cycle 3 (`#785` closed earlier today; `#769` and `#798` remain).

sandeepkunkunuru:
Correction to my previous comment: **this PR is not rebased.** Its head is on a fork, so the branch I pushed and verified landed in this repo as a separate ref rather than updating this PR. Its head is still `408c20f`, eleven merges behind `main`, and no CI has run on it.

The verified rebase is at #842 — same diff, replayed onto `5c8b177`. I have left this PR alone rather than force-pushing to someone else's fork.

56steve:
Closing as superseded — the identical fix (`carry_names_through_with`, applied in both AST shapes) has since landed on `main` via #842, which carries the same title. Thanks! The behaviour this addressed is now covered upstream.

Issue #855, opened 2026-08-25
Lists have no ordering arm at all, and NaN orders as null instead of false


sandeepkunkunuru:
## 1. Lists never order

```
RETURN [1, 0] >= [1]      expected true    got null
RETURN [1, 2] >= [3, null]  expected false  got null
RETURN [1, null] >= [1]   expected true    got null
```

`cypher_ordering` had **no arm for `Array`**, so every list comparison fell to `_ => None` and answered null. Lists order lexicographically and then by length: `[1, 0] >= [1]` is true because the shared prefix is equal and the left is longer.

A null at a position that actually has to be compared makes the answer undecidable, which is what `None` means. It is not reached by a list that is merely longer: `[1, null] >= [1]` is decided by the length and the null is never compared.

## 2. NaN orders as null instead of false

```
WITH 0.0 / 0.0 AS nan RETURN nan > 1, nan >= 1, nan < 1, nan <= 1
  expected false, false, false, false      got null, null, null, null
```

`partial_cmp` returns `None` for NaN, and this branch maps `None` to null — the same answer it gives for two values that cannot be compared at all. Cypher separates them: incomparable is null, NaN is false, and all four operators are false, `>=` against itself included.

**Only against another number.** `0.0/0.0 < 'a'` is still null, because comparing across types is null before NaN is considered — returning false there would claim "I compared them and they did not order", which is a different statement. A blanket NaN rule cost exactly that one scenario, and the regression diff named it.

## Effect

TCK +11, zero regressions. `Comparison2` goes to 0 failures.

## Not fixed

`Return2` scenario 18 wants an **unknown function** to raise a `SyntaxError` at compile time; the engine raises it at runtime, and the scenario's empty graph means the call is never reached. That is the same shape as #843, but the fix needs the validator to know every function name the dispatcher accepts — and a name it fails to list becomes a *rejected valid query*, which is the worse failure. Left open deliberately rather than guessed at.

Issue #840, opened 2026-08-25
A WITH's WHERE cannot see the variables the projection drops, and ordering an entity raises instead of yielding null


sandeepkunkunuru:
Two independent defects that had to be fixed together, because each alone leaves the same scenario failing.

## 1. `WITH … WHERE` loses the pre-projection scope

```cypher
UNWIND [0,1] AS i UNWIND [0,1] AS j
WITH i AS a, j AS b
WHERE i <> j
RETURN a, b
```

Two rows expected, **zero returned** — no error. The predicate is applied *inside* the barrier, after the projection has dropped `i` and `j`, so it evaluates to null for every row. A filter that matches nothing is a legitimate outcome, so nothing distinguishes this from a correct empty result.

Neo4j answers the `Comparison2` scenario that uses this shape (its recorded actuals return `[["1", "3.14"]]`), so the reference behaviour is that those names are still reachable.

**Fix:** split the predicate by conjunct. A conjunct moves ahead of the barrier only when it names at least one variable and none of them is a projected alias — so `WITH n.name AS name WHERE name STARTS WITH 'a'` still runs after, and an aggregate's output can never be filtered before it is computed.

The split belongs in `build_with_barrier` because **both** planner paths call it. The by-kind stage loop and the clause pipeline each build their own WITH stages, and my first attempt patched the stage loop — which moved nothing, because the query in question parses into the pipeline. Same trap as #797.

## 2. Ordering an entity raises instead of yielding null

```
MATCH (n) RETURN n < 1
-- TypeError("Binary op requires property values")
```

Cypher's rule is that comparing across types yields null except between numbers, and a node, relationship or path is simply another type that does not order. Raising took down the whole query: `Comparison2` scenario 3 builds a list holding one value per type and compares every pair, so one `TypeError` on `node < ''` lost all 90 rows — including the numeric pair the scenario is actually about.

Only the ordering operators. `=` and `<>` on entities are identity and already handled; arithmetic on an entity stays an error, because there a null really would hide a mistake.

## Why together

Fixing either alone gains **nothing measurable**: the comparison fix leaves the scenario at zero rows from the scope bug, and the scope fix leaves it erroring on the first entity pair. I measured the comparison fix on its own and it was +0 with zero regressions, which is what sent me looking for the second half.

## Effect

TCK +8, zero regressions. `Comparison2` 10 failures to 6; the remainder are lexicographic list ordering, a separate rule.

PR #861, opened 2026-08-25
One comparison engine, and it knows about entities and numbers (#860)

### LINKED ISSUE #860: 1 = 1.0 is false, relationship and path equality raise, and a MATCH's WHERE uses a second comparison engine
Three defects that turn out to be one story: **a `WHERE` attached to a `MATCH` was using a different, weaker comparison implementation than the same expression anywhere else.**

## 1. `1 = 1.0` is false

```
RETURN 1 = 1.0     expected true    got false
```

`cypher_equals` ended in `_ => Some(a == b)`, and `PropertyValue`'s derived `PartialEq` compares *variants* — so an integer never equalled a float. A wrong answer in the most ordinary comparison there is, and one that reads as a legitimate "these differ".

Now compared **as integers**, not through `as f64`: a float with a fractional part cannot equal an integer, and one without is exact, so no bit is lost above 2^53.

## 2. Relationship and path equality raise

```
MATCH ()-[r]->() RETURN r = r
-- TypeError("Binary op requires property values")
```

Only *nodes* had an identity arm. Relationships, paths, and an entity compared against a non-entity all fell through to the property-value requirement.

Paths compare structurally — same nodes, same relationships, same order. The TCK's scenario is a **self-loop**, where a path and its reverse have the identical node sequence, so a direction-insensitive rule would be inventing something it does not test.

## 3. The real cause: two comparison engines

```
MATCH ()-[a]->() MATCH ()-[b]->() WHERE a = b     TypeError
MATCH ()-[a]->() MATCH ()-[b]->() RETURN a = b    true
```

`FilterOperator::evaluate_binary_op` was a second implementation — **67 lines against the free function's 346** — and its own comment already said so:

> *Note this is a **second** comparison implementation — the two must agree, and did not.*

It agreed on the easy things and diverged on every rule added since. It was missing relationship and path identity, the three-valued list and map equality of `cypher_equals`, the NaN and list-ordering rules from #855, the entity-ordering rule from #840, and integer-float equality — so **every comparison rule fixed this cycle applied everywhere except the clause most queries actually filter in.**

It already delegated `^` and `XOR` "so they have one definition rather than two that can drift". Now it delegates everything.

## Effect

TCK +3 across the three, zero regressions. The count understates it: the scenarios that exercise these rules mostly do so in a `RETURN`, where they already worked.

sandeepkunkunuru:
Closes #860. Three defects, one story: **a `WHERE` attached to a `MATCH` was using a different, weaker comparison implementation than the same expression anywhere else.**

```
MATCH ()-[a]->() MATCH ()-[b]->() WHERE a = b     TypeError
MATCH ()-[a]->() MATCH ()-[b]->() RETURN a = b    true
```

`FilterOperator::evaluate_binary_op` was a second implementation — **67 lines against the free function's 346** — and its own comment already said so:

> *Note this is a **second** comparison implementation — the two must agree, and did not.*

It agreed on the easy things and diverged on every rule added since: relationship and path identity, `cypher_equals`' three-valued list and map equality, the NaN and list-ordering rules (#855), the entity-ordering rule (#840). **Every comparison rule fixed this cycle applied everywhere except the clause most queries actually filter in.** It already delegated `^` and `XOR` "so they have one definition rather than two that can drift"; now it delegates everything.

## `1 = 1.0` was false

`cypher_equals` ended in `_ => Some(a == b)`, and `PropertyValue`'s derived `PartialEq` compares *variants* — so an integer never equalled a float. Compared **as integers** now, not through `as f64`: a float with a fractional part cannot equal an integer, and one without is exact, so no bit is lost above 2^53.

## Relationships and paths had no identity arm

Only nodes did. Paths compare structurally — the TCK's scenario is a **self-loop**, where a path and its reverse have the identical node sequence, so a direction-insensitive rule would be inventing something it does not test.

## The parity test was rewritten after it failed to earn its keep

I wrote it first over property values — `1 = 1.0`, list equality, NaN — and **it passed against the old engine too**, because that engine's scalar helpers had kept up. What it had never gained was the entity arms.

It now compares *entity* predicates across `MATCH … WHERE`, `WITH … WHERE` and `RETURN`, and I verified it **fails against the reverted code** rather than assuming it would. The scalar cases are kept and labelled as guarding a future divergence rather than as having caught this one.

## Verification

- TCK **3,630 → 3,633**, regression diff against the merge base **empty**. The count understates it: the scenarios that exercise these rules mostly do so in a `RETURN`, where they already worked.
- `--min-pass` ratcheted 3630 → 3633.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,7 +134,7 @@ jobs:
         # why in the same commit.
         run: |
           git clone --depth 1 --branch 2024.3             https://github.com/opencypher/openCypher.git /tmp/openCypher
-          cargo run --release --example tck_runner --             --features /tmp/openCypher/tck/features --min-pass 3630
+          cargo run --release --example tck_runner --             --features /tmp/openCypher/tck/features --min-pass 3637
 
       - name: CH-DETERM-THREADS — the answer does not depend on the thread count
         # LANG-14 (#611). CH-DETERM varies the *process*, which catches a
--- a/src/query/executor/mod.rs
+++ b/src/query/executor/mod.rs
@@ -1623,11 +1623,23 @@ mod tests {
                 node.set_property("active", PropertyValue::Boolean(i % 2 == 0));
             }
         }
-        // Query: compare Boolean property to String literal 'true'
+        // Comparing a Boolean property to the *string* 'true' matches nothing.
+        //
+        // This asserted 2 rows, on the strength of a String/Boolean coercion in
+        // `coerced_eq` that Cypher does not have. Comparing across types is
+        // false: `1 = '1'` is false for the same reason, which the TCK settles
+        // in `Comparison1` scenario 9. The test encoded the coercion rather than
+        // the language, and was corrected against the reference when the second
+        // comparison engine that provided it was deleted (#860).
         let executor = QueryExecutor::new(&store);
         let query = parse_query("MATCH (i:Item) WHERE i.active = 'true' RETURN i.name").unwrap();
         let result = executor.execute(&query).unwrap();
-        // Item0=true, Item1=false, Item2=true, Item3=false
+        assert_eq!(result.records.len(), 0, "a boolean never equals a string");
+
+        // The same property compared to a boolean still works, so this is not
+        // simply everything matching nothing.
+        let query = parse_query("MATCH (i:Item) WHERE i.active = true RETURN i.name").unwrap();
+        let result = executor.execute(&query).unwrap();
         assert_eq!(result.records.len(), 2, "Expected 2 active items");
     }
 
--- a/src/query/executor/operator.rs
+++ b/src/query/executor/operator.rs
@@ -160,6 +160,22 @@ fn cypher_equals(a: &PropertyValue, b: &PropertyValue) -> Option<bool> {
             }
             if unknown { None } else { Some(true) }
         }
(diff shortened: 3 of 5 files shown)


sandeepkunkunuru:
Force-pushed with the fix for what CI caught, plus #862 folded in.

**CI found a real defect in this change and I had not.** `test_cross_type_string_boolean_eq` asserted that `i.active = 'true'` matches the boolean `true` — on the strength of a String/Boolean coercion in `coerced_eq` that **Cypher does not have**. `1 = '1'` is false for the same reason, which the TCK settles in `Comparison1` scenario 9. The test encoded the coercion rather than the language; it is corrected against the reference, with a second assertion that comparing to a real boolean still matches two rows, so it cannot pass by matching nothing.

I had run the targeted tests and the TCK before pushing but not the workspace suite. That is what found it.

Also in this push:

- **The seventeen dead helpers are deleted, not left unused.** They were the body of the second engine; keeping them invites the next change to route through them again, which is how the two drifted apart.
- **#862 folded in**: `weekDay` returned null, because only the constructor's `dayOfWeek` spelling was in `temporal_component`. Every other date and date-time accessor was checked against the TCK's expected values and already agreed. +4.

Re-verified: TCK **3,630 → 3,637**, regression diff empty; lib **2,175 tests**; workspace **165 suites / 3,367 tests**; all green.

PR #1091, opened 2026-09-04
The cyclic prune handles directed closes, which is what FinBench CR-4 is

### LINKED ISSUE #1090: Widen the cyclic prune to directed closes — the blocker is which half of the type index gets built
#1088 prunes a cyclic pattern's middle hop against the node the next hop closes onto, and takes BI-17 from 6,874 ms to 872 ms at SF1. It fires only when **both segments are undirected** — deliberately narrow, so that "neighbours the closing variable" has one unambiguous meaning.

Widening it to directed closes is mostly bookkeeping, and there is one detail that is not.

### The easy part

The closing segment's source is this expand's target (`c`) and its target is the bound variable (`a`), so:

| closing segment | `c` must be … | list to search |
|---|---|---|
| `(c)-[:R]->(a)` | an in-neighbour of `a` | `in_of(a)` |
| `(c)<-[:R]-(a)` | an out-neighbour of `a` | `out_of(a)` |
| `(c)-[:R]-(a)` | either | both (what #1088 does) |

That is a parameter on `with_co_neighbour`, and the planner already knows `next.edge.direction`.

### The part that is not

`co_lists` is built from `typed`, the type index this expand acquired — and **which halves it holds depends on the middle expand's own direction**, not on the close's:

```rust
let out = store.type_adjacency(t, true);
let inc = match self.direction {
    Direction::Outgoing => None,          // <- never fetched
    _ => store.type_adjacency(t, false),
};
```

So `MATCH (a)-[:R]->(b)-[:R]->(c)-[:R]->(a)` — an Outgoing middle segment with a close needing `in_of(a)` — would find `inc` is `None`, `in_of` returns the static empty slice, and the test would be silently skipped. Correct, and worth nothing: exactly the shape where a widening lands, looks implemented, and does not run.

So the change is not "pass the direction through". It is: **the index halves a walk needs and the halves a close needs are two different questions, and the acquisition currently answers only the first.** Either fetch both halves when a co-neighbour var is set, or make acquisition take the union of what both need.

### Worth checking first

Whether any benchmark query has a *directed* cyclic close at all. BI-17 is undirected; if nothing in SNB, BI or FinBench closes directionally, this is speculative work and the narrow version is the right amount. `examples/bi17_plan.rs` prints plans, and the four benches' query sets are the population to grep.

That check is cheap and should come before the implementation — the last four things tried on BI-17 were each ruled out by a measurement that took minutes.

sandeepkunkunuru:
Closes #1090.

#1088 refused any close that was not undirected. **FinBench CR-4 is exactly the shape it refused:**

```cypher
MATCH (a:Account {id: $A1})-[t1:TRANSFER]->(b:Account)
      -[t2:TRANSFER]->(c:Account)-[t3:TRANSFER]->(a)
```

Structurally BI-17 with every hop directed — and a directed transfer cycle is the money-laundering motif behind the fintech and LEA work, not an incidental benchmark row.

### The blocker was not the direction, it was the index halves

Acquisition asked only what the **walk** needs:

```rust
let inc = match self.direction {
    Direction::Outgoing => None,        // <- CR-4's middle hop lands here
    _ => store.type_adjacency(t, false),
};
```

CR-4's middle hop is `Outgoing` and its close needs `a`'s **incoming** list. A widening that only threaded the direction through would have compiled, looked implemented, and **never run on the query that motivated it**. The requirement is now the union of what the walk needs and what the close needs.

Usability is still judged on the walk alone: the close is an optimisation whose absence costs rows and never answers, so a missing close-half must not disable the walk's fast path.

### Verified by mutation

Swapping the two arms of the direction mapping turns `a_directed_close_answers_the_same_as_a_separate_match` **red** while every other test in the file stays green — so the prune genuinely fires on directed closes, and searching the correct half is what makes the answer right.

The undirected tests cannot catch it: they search both lists, which makes the two arms indistinguishable. `a_directed_close_searches_the_incoming_list_not_the_outgoing_one` pins it on a graph whose in- and out-neighbourhoods are disjoint.

### Unchanged

BI-17: 0.887 s, 387,573 triangles. 7 cyclic tests, 11 `type_adjacency` tests.

The CR-4 win should be **measured on FinBench SF10, not assumed from BI-17** — CR-4 anchors on a single account where BI-17 scans every person, so the fan-out and the row saving are much smaller.

https://claude.ai/code/session_01S5sAcU4QgoPoZP3FgR3Nbw

### DIFF
--- a/src/query/executor/operator.rs
+++ b/src/query/executor/operator.rs
@@ -6264,6 +6264,18 @@ pub struct ExpandOperator {
     /// it only when both segments are undirected over the same edge type, so
     /// "is a neighbour" has one unambiguous meaning.
     co_neighbour_var: Option<Arc<str>>,
+    /// Which of the closing variable's neighbour lists proves the close.
+    ///
+    /// The closing segment runs `target -> co_neighbour_var`, so an
+    /// `Outgoing` close means the target must be an *incoming* neighbour of
+    /// that variable, and vice versa. `Both` searches either.
+    ///
+    /// It has to be carried separately from `direction`, which is *this*
+    /// expand's. FinBench CR-4 --
+    /// `(a)-[:TRANSFER]->(b)-[:TRANSFER]->(c)-[:TRANSFER]->(a)` -- has an
+    /// outgoing middle hop and needs `a`'s incoming list; reading `direction`
+    /// would search the wrong one (#1090).
+    co_neighbour_dir: Direction,
 }
 
 impl ExpandOperator {
@@ -6292,6 +6304,7 @@ impl ExpandOperator {
             track_edges: false,
             starts_clause: false,
             co_neighbour_var: None,
+            co_neighbour_dir: Direction::Both,
             target_bound_var: None,
             direction,
             current_record: None,
@@ -6331,9 +6344,11 @@ impl ExpandOperator {
         self
     }
 
-    /// Keep only targets that also neighbour `var`. See `co_neighbour_var`.
-    pub fn with_co_neighbour(mut self, var: String) -> Self {
+    /// Keep only targets that also neighbour `var`, in the direction the
+    /// closing segment requires. See `co_neighbour_var`.
+    pub fn with_co_neighbour(mut self, var: String, closing: Direction) -> Self {
         self.co_neighbour_var = Some(var.into());
+        self.co_neighbour_dir = closing;
         self
     }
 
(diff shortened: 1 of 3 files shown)


sandeepkunkunuru:
## CR-4 is 8.3× slower than Neo4j at SF10 today, so this has a measurable target

From the 2026-09-01 three-engine FinBench SF10 run:

| engine | CR-4 Transfer Cycle Detection |
|---|---:|
| **Samyama** | **541 ms** |
| Neo4j | 64.8 ms |
| FalkorDB | 90.8 ms |

**8.3× the best competitor**, on a query whose name is literally "Transfer Cycle Detection". PERF-03 asks FinBench within 2× at SF10, so CR-4 is one of the queries missing that bar, and it is the exact shape this PR stops refusing.

The prune will fire on it: segments are `a→b`, `b→c`, then `c→a` closing onto the bound `a`, same `TRANSFER` type throughout, neither variable-length. What it needs is `a`'s incoming list against an `Outgoing` middle hop — the union fix above.

### What I am not claiming

CR-4 anchors on **one** account (`{id: $A1}`) where BI-17 scans every person, so the fan-out and therefore the row saving are much smaller. 541 ms → under 130 ms would clear the 2× bar; I would not project a 7.88×, and I have not measured it — there is no FinBench extract on this host.

It also binds `t1`, `t2`, `t3`, so relationship-isomorphism bookkeeping is on throughout, which the BI-17 measurements did not exercise.

**A FinBench SF10 run would settle two things at once**: this number, and samyama-graph#1048 (CR-11 returns 3 rows where Neo4j and FalkorDB both return 1) — the strongest open P0 candidate, which is what H1 condition 5's triage is blocked on (samyama-graph-competitor-benchmarks#133).

Issue #1183, opened 2026-09-09
v1.7.1 exhausts memory on a 2-node graph; fixed on main but the release is exposed


sandeepkunkunuru:
## v1.7.1 exhausts memory on a two-node graph

`samyama v1.7.1` — the current release — grows without bound on a variable-length pattern over a graph with **2 nodes and 3 edges**. Measured peak **75 GB** during the GPML conformance suite; it repeatedly killed a 46 GB workstation, taking the terminal session with it.

**Already fixed on `main`.** This issue is about the released artifact, not about work still to do.

## Reproducer

Graph — two nodes, three edges, one of them a self-loop:

```
(x:N)-[:E]->(x)     e1   self-loop
(x:N)-[:E]->(y:N)   e2
(y:N)-[:E]->(x:N)   e3
```

Query:

```cypher
MATCH p=(x:N)-[:E*1..2]->(y:N) RETURN x.eid AS s, y.eid AS t
```

| build | result |
|---|---|
| **v1.7.1** (released 2026-08-27) | exceeds a **6 GB** cap; peaked at **75 GB** under the full suite |
| **main** `717cde3` | **4 rows, 0.0 s, 20 MB peak** |

Under TRAIL, `*1..2` on three edges has a handful of paths. There is no input here large enough to justify any of this.

## Why it went unnoticed

The workstation deaths read as the *session* being killed, not a query — nothing in the engine logged anything, and `mem-guard` never fired because the process outran its sampling. Running the suite on a 128 GB AWS spot box is what made the shape visible: `samyama-v171` climbing 23.5 GB → 53 GB → 75 GB while every other process stayed under 800 MB.

It is not exercised by our own tests. The corpus that finds it is the metamorphic layer of an external suite, which asks quantifier-range decompositions (`{0,4}` against `{0,0}+{1,1}+…`) over deliberately pathological small fixtures — self-loops and 2-cycles. Our fixtures do not contain those shapes at those bounds.

## What this changes

The gap between `main` and the release is no longer only a conformance-score gap. On the released engine this is a **denial of service reachable from an ordinary read query** on a trivially small graph — relevant to any instance serving untrusted or generated Cypher, and to the NLQ path in particular.

161 commits sit on `main` unreleased, including #1144, #1145, #1146, #1147 and #1149. Whatever the release cadence, the argument for cutting one is now stronger than "the conformance table looks better".

## Not yet determined

[...]

Issue #750, opened 2026-08-24
A string property read allocates: 69% of IS3 is malloc, four String clones per row


sandeepkunkunuru:
**Profiled, not inferred: 69% of LDBC IS3 is the allocator, and the allocations
are `String` property clones — roughly four per row.**

This is the fourth attempt at IS3 on #718 and the first with a profile behind
it. Every earlier candidate (property-access cost, relationship binding, the
client result path, position-in-suite, the adjacency tier) was measured and
ruled out.

## The profile

`perf record -e cpu-clock -D 40000` — the delay skips the 30 s load so only the
query phase is sampled. SF1, p95-degree anchor (144 friends, matching SF10's
137), 250,000 executions:

| symbol | share |
|---|---:|
| `_int_malloc` | **45.1%** |
| `malloc_consolidate` | 8.0% |
| `unlink_chunk` | 6.1% |
| `cfree` | 4.9% |
| `_int_free` | 2.9% |
| `malloc` | 1.8% |
| **allocator total** | **≈69%** |
| `edge_type_matches` | 2.3% |
| `ProjectOperator::next_batch` | 1.9% |
| `quicksort` | 1.8% |

## Which allocations

Integer against string at identical structure and row count:

| variant | median | delta per row |
|---|---:|---:|
| project 1 **int** column | 0.046 ms | — |
| project 1 **string** column | 0.066 ms | **+139 ns** |
| project 1 int + `ORDER BY` that int | 0.057 ms | +76 ns |
| project 1 string + `ORDER BY` that string | 0.104 ms | **+264 ns** |

A `String` property costs ~139 ns more than an integer to project and ~264 ns
more as a sort key. `PropertyValue::String(String)`, so **every read of a string
property allocates and copies**.

IS3 touches four strings per row — `firstName` and `lastName` as sort keys, then
both again in the projection — which at those rates is 86–150 µs against IS3's
measured 153 µs. That is the whole query.

## Two fixes, and why I am filing rather than building

**1. Stop reading the same property twice (#593).** IS3's plan is

```
Project (f.id, f.firstName, f.lastName)
+- Sort (f.firstName ASC, f.lastName ASC)
```

Sort clones both strings for its keys and Project clones both again. Removing
that duplication takes four string touches per row to two — **~40% of IS3** on
these numbers. #593 measured this as "6 µs on 23 rows" and reasonably set it
aside; at 144 rows with two shared *string* columns it is the largest single
[...]

Issue #602, opened 2026-08-18
[PERF-10] set_edge_property writes every value three times — 728 B/edge of duplication, and edge properties cost 2x node properties


sandeepkunkunuru:
## Measured

`benches/prop_scaling.rs` (added in the accompanying PR), 200K nodes, degree 3, release, `389844e`, Linux x86_64 16 vCPU:

| node props | edge props | B/node | B/edge (marginal) |
|---:|---:|---:|---:|
| 0 | 0 | 365.0 | 134.3 |
| 2 | 0 | 781.3 | 134.3 |
| 4 | 0 | 1159.5 | 134.3 |
| 8 | 0 | 1938.1 | 134.3 |
| 8 | 3 | 1938.1 | **1332.2** |

**~197 B per node property. ~399 B per edge property** — twice the node cost for the same value. A bare edge is **134.3 B**, at PERF-10's 128 B H2 target; three properties multiply it by **9.9×**.

## The duplication

Same fixture, edge properties written through `set_edge_property_sparse` (row map only):

| write path | B/edge, 3 properties |
|---|---:|
| `set_edge_property` | 1332.2 |
| `set_edge_property_sparse` | **603.9** |

**728 B/edge of duplication — 243 B per property, 55% of the cost.**

`GraphStore::set_edge_property` (`src/graph/store.rs:1378`) writes the same value to three places:

```rust
self.edge_columns.set_property(idx, &key_str, val.clone());   // columnar
self.set_edge_property_sparse(edge_id, key_str, val);         // row map
...
let post_props = self.edge_properties.get(&edge_id).cloned()  // full clone
    .unwrap_or_default();
version_log.push(/* ... post_props ... */);                   // MVCC version log
```

This is **#545's defect class on the edge side, at 3.4× the per-object size** (#545 measures 217 B/node). Nodes carry two copies; edges carry three.

## Why this reframes PERF-10

`memory_footprint.rs` fixes the fixture at two node properties and zero edge properties — LDBC's shape. That is a reasonable default and it is also why PERF-10 reads as a bytes-per-edge problem: the harness holds constant the term that varies most.

Adding the measured terms gives

```
resident ≈ 1.10 × [ N × (365 + 197 × node_props) + E × (134 + 399 × edge_props) ]
```

which predicts **real LDBC SF1 at ~10.9 GB against the measured 9,565 MB — within 15%**, from two synthetic runs and a property count. A single bytes/edge constant has no term for what varies and cannot do this.

[...]

Issue #491, opened 2026-08-16
[PERF-10/PERF-14] 19.5 heap allocations per edge and 9 per node — compact storage was added beside the old representation, not instead of it


sandeepkunkunuru:
## Measurement

The footprint harness (#490) counts allocation **calls**, not just bytes. At 100k nodes, degree 4:

| Phase | allocations | per object |
|---|---:|---:|
| bare nodes (no properties) | 900,069 | **9.00 / node** |
| node properties (2 per node) | 1,900,035 | **19.00 / node** — 9.5 per property |
| edges | 7,800,538 | **19.50 / edge** |

Exact integers, so these are deterministic per-object counts rather than measurement noise.

**Creating one edge takes 19.5 trips to the allocator.** It should take one or two.

## Where they come from

`GraphStore::create_edge` (`src/graph/store.rs:1448`), reading it against the counter:

```rust
let edge_type = edge_type.into();                       // 1: allocate EdgeType(String)
let mut edge = Edge::new(edge_id, source, target, edge_type.clone());   // 2: clone it
...
self.edge_type_index.entry(edge_type.clone())           // 3: clone it again
    .or_insert_with(HashSet::new).insert(edge_id);
...
let src_labels: Vec<Label> = self.get_node(source)
    .map(|n| n.labels.iter().cloned().collect())        // 4: Vec + a String clone per label
    .unwrap_or_default();
let tgt_labels: Vec<Label> = ...                        // 5: and again for the target
self.catalog.on_edge_created(source, &src_labels, &edge_type, target, &tgt_labels);
```

The same edge-type string is allocated and cloned **three times**, and two `Vec<Label>` are built with cloned `String`s on **every edge insert** purely to hand the catalog some labels it could have looked up itself.

## The structural point

The compact representations already exist. `create_edge` maintains:

- `edge_endpoints[idx] = (source, target)` — DS-07c
- `edge_type_ids[idx] = intern_edge_type(&edge_type)` — interned, documented in `create_edge_stub` as *"Compact edge type: 2 bytes per edge (DS-07c)"*

…**and also** the old `Edge` struct carrying `EdgeType(String)`, and `edge_type_index: HashMap<EdgeType, HashSet<EdgeId>>` keyed by the string form.

So DS-07c was added **beside** the legacy representation rather than replacing it, and we pay for both. The interning work that #477 proposed doing is in the tree already — it just is not the only copy.

[...]

PR #1184, opened 2026-09-09
release: v1.8.0

sandeepkunkunuru:
Cuts **v1.8.0**. Merging this fires `tag-release.yml`, which creates the tag and the GitHub Release, which in turn fires the crates.io / PyPI / npm / Docker publishes.

## Why now, and not on cadence

`gpml-conformance` measures **releases a reader can install**. Ours is v1.7.1 from 2026-08-27, which predates every fix that suite prompted — so our published engine is the worst row in a study we wrote:

| build | conforms | diverges | rejects | inexpressible | S1 | S2 |
|---|---:|---:|---:|---:|---:|---:|
| v1.7.1 released | 8 | 4 | 0 | 5 | 0.33 | 1.00 |
| main `717cde3` | **15** | 2 | 0 | 0 | **0.12** | 1.00 |
| neo4j-2026 | 14 | 2 | 1 | 0 | 0.12 | 0.67 |

A leadership claim resting on an unreleased build invites exactly the objection the artifact exists to survive.

**And the released engine has a worse problem than the conformance gap.** #1183: v1.7.1 grows without bound on a graph of **2 nodes and 3 edges** — a self-loop plus a 2-cycle asked `*1..2`, which under TRAIL has four rows. Peak **75 GB**; `main` answers in **20 MB**. On the released engine that is a denial of service reachable from an ordinary read query.

Rationale recorded in samyama-cloud `wiki/decisions/v180-release-why-and-how.md`.

## Two things this PR fixes beyond the bump

**1. `prepare-release.yml` could not open its own PR.** It pushed `release/v1.8.0` correctly and then failed with *"GitHub Actions is not permitted to create or approve pull requests"* — an org setting. The branch was fine; only the last step failed. This PR was opened by hand. Either that org setting needs changing or the workflow should stop claiming to open a PR.

**2. The bump missed three version locations, and the repo's own test caught it.** Six manifests moved; `api/openapi.yaml` still said 1.7.1 in **three** places, as did prose in `src/lib.rs` and `CLAUDE.md`. `tests/version_consistency.rs` guards exactly those, so this PR would have gone red on its own consistency check.

The guard worked. The bump did not. `prepare-release.yml` now bumps them too — deriving the previous version from `Cargo.toml` rather than trusting an input, then grepping to confirm nothing survives, because **a `sed` that matches nothing is silent**, which is how this was missed.

`openapi.yaml` is the one that matters: as its own test comment puts it, *"an example is what a reader copies, and a status example claiming 0.7.0 is a published claim about the server."*

## Verification

All 7 `version_consistency` tests pass. CI on this PR is the real check.

## After merge

Re-measure `gpml-conformance` against the v1.8.0 release so the paper's row is a build anyone can install, then update the paper. The claim will be narrow and exact: **the only engine that answers all 17 constructs, ahead of Neo4j 2026 by one construct** (`mode-simple-bounded`, which they reject), with the same two divergences — and **worse on S2**, 1.00 against their 0.67.

## Update 2026-09-14 — rebased; ten internal pins were still 1.7.1

**Not for merge yet: held for team review.**

Rebased on main `0eb6a55`. The release now also ships:
- #1186 (edge property reads without cloning),
- #1193 (FinBench loaders write relationship properties to columns),
- #1196 (typed DateTime column; IC2 0.73× Neo4j, about −4 GB on SNB SF10).

**3. The bump also missed every internal dependency pin.** Crates depend on each other by path *and* version, for example `samyama-sdk = { path = "../crates/samyama-sdk", version = "1.7.1" }`. All ten pins (root ×4, `cli`, `sdk/python`, `samyama-sdk` ×3, `samyama-graph-algorithms`) still named 1.7.1. It builds, because 1.8.0 satisfies `^1.7.1`. So nothing noticed. But the published crates would each accept any 1.7.x-or-later sibling from crates.io.

- `175818d`: `every_internal_dependency_pin_names_the_version_of_record` in `tests/version_consistency.rs`. It fails on the branch as it stood. It also asserts that it found at least 5 pins, so an empty scan can't pass.
- `1e01a42`: bumps the ten pins. `prepare-release.yml` gets a *Bump internal dependency pins* step that seds `path = "…", version = "PREV"` in every manifest, then greps to confirm none survive. Dry-run on a copy, 1.8.0 → 1.8.1: 10 pins moved, the package version line untouched.
- The previous version is now read from `HEAD:Cargo.toml` instead of `HEAD~1`. `HEAD~1` names an older version whenever main's tip is itself a version change.

## Update 2026-09-14 (2): rebased on main after #1198 and #1194

Rebased onto main `7f75a28`, so v1.8.0 also ships:

- **#1198**: footprint and latency.
  - The variable-length expand reads the per-type index (#1197).
  - `set_edge_property` keeps no property-map copy (#602).
  - Creating a relationship takes 1 allocator call instead of 5 (#491).
  - Relationship references no longer copy properties (#1190).
  - `ORDER BY` no longer copies (#750).
- **#1194**: the LDBC loaders and the banking demo write relationship properties to columns.

Measured on LDBC SNB SF10 on one host (Vultr, 16 dedicated vCPU / 128 GB), median of 5 runs, 21/21 queries passing on each build:

| | main before (`0eb6a55`) | #1198 | #1194 |
|---|---:|---:|---:|
| IC1 | 721 ms | **55.7 ms** | 667 ms |
| IS3 | 0.54 ms | 0.45 ms | 0.50 ms |
| IC11 | 74.7 ms | 63.9 ms | 67.6 ms |
| peak RSS | 93.49 GB | 93.49 GB | 93.89 GB |

- The first IC1 after a load or a write builds the `KNOWS` type index, which cost 671 ms in the profile run.
- #1198 moved the heavy queries (IC3/5/6/9/10/12) by 1.03–1.07×, under CH-REGRESS's 1.1× + 5 ms.
- Both PRs were measured separately against the old main. The combined build has not been benchmarked.

The four release commits are unchanged: the version bump, `prepare-release.yml` for the non-manifest locations, and the ten internal dependency pins with their test.


sandeepkunkunuru:
**Held for a fresh CH-REGRESS verdict** (2026-09-14, owner's decision).

Rebased onto main `7f75a28` (#1198 and #1194). The rebase was clean, and nothing version-guarded still names 1.7.1.

The release gate would pass today, but on the wrong evidence:

- `RELEASE-GATE.json` says `pass`, 2 days old (limit 3). It was measured on engine `0eb6a55`, before #1198 and #1194.
- `check_release_gate.py` checks status and age, not the engine commit, so v1.8.0 would pass a regression check that never ran on what it ships.
- SNB SF10 was measured on the new code (heavy IC queries 1.03–1.07×, within 1.1× + 5 ms). BI and FinBench were not.
- The spot nightly that would refresh the verdict has failed since 09-12: first the AWS keys were rejected, now the account is blocked for `RunInstances`. A non-AWS host doesn't help, because the baseline refuses any host other than `aws-r6a.4xlarge-spot`.

After about 21:56 UTC on 09-15 the gate refuses the tag regardless, until a new verdict is published.

**To unblock:** the AWS account is cleared, and the nightly (21:40 UTC) builds main's tip and publishes a verdict for it. Then this PR merges.

Worth considering separately: `check_release_gate.py` could also require `engine_commit` to be an ancestor of the commit being tagged, no more than N commits behind it. Today a verdict from any older main satisfies it.

PR #1189, opened 2026-09-10
The release gate demanded a status CH-REGRESS never emits

Kalyanpulikanti-01:
No tag has been cut since **2026-08-28**, the day `scripts/check_release_gate.py` landed. Not because anything regressed — because the condition could not hold.

## The bug

```python
if g.get("status") != "pass":
    return 1
```

The envelope's status is derived pessimistically (benchmarks repo, `harness/envelope.py`):

```
error      if any measurement errored
fail       if any measurement failed        <- a real regression lands here
partial    if any measurement is partial or unmeasured
pass       otherwise
```

CH-REGRESS marks each of its **62 per-query timings `partial`**, because a timing has no pass/fail target of its own. So the run-level status is `partial` however green the suite is, and `pass` is unreachable.

## Evidence it is categorical, not selective

| | |
|---|---|
| verdicts published since 2026-08-31 | 12 — all `partial` or `fail`, never once `pass` |
| this check landed | `4746793`, 2026-08-28 |
| release tags containing that commit | **0** |
| v1.7.1 tagged | 2026-08-27 — one day earlier, which is why it escaped |

v1.7.1 is the last installable release *because* it predates this check. Everything since has been stuck behind a gate that has never opened — including the fix for the unbounded-memory blow-up (#1183) that is live in what users install today.

## The benchmarks repo already named this trap

`harness/gate.py`, in the docstring for `_newest_envelope_status`:

> *"testing for `"pass"` would be a condition that can never hold however green the suite is … That is the third check today built so it could not pass; the other two were in the quiet-host sampler. It is an easy shape to write and an **invisible one to read**, because 'not met' looks like work remaining rather than like a broken test."*

That file uses the right definition. This one didn't.

## The fix

Accept `partial` alongside `pass`.

**This does not weaken the gate against regressions.** By the rollup above, a slower query yields `fail` — still refused, as are `error` and `unmeasured`. Staleness and a missing verdict still block exactly as before.

## The test was the other half

`scripts/test_check_release_gate.py` covered `pass`, `fail`, `unmeasured`, stale, a missing verdict, and the day-3/day-4 boundary. Thorough enough to look reassuring — and it **never tested `partial`**, the one status the suite actually produces. It passed in CI throughout.

The new case fails against the unfixed script, which is the point of adding it:

```
$ git stash -- scripts/check_release_gate.py    # revert to the original
$ python3 scripts/test_check_release_gate.py
...
FAIL
```

## Verified against the live verdict

Against the file currently on the `release-gate` branch:

```
$ python3 scripts/check_release_gate.py RELEASE-GATE.json
CH-REGRESS partial, 0d old, host 'vm-1', engine e728b28
$ echo $?
0
```

So this unblocks the tag **with no new measurement** and nothing needed on `vm-1`.

## One gap left open deliberately

`partial` also covers *"a measurement declined to judge"*, and the verdict file carries no per-measurement detail to tell that apart from the ordinary timing case. `harness/nightly/publish_verdict.py` reduces the envelope to seven fields.

Closing it needs the nightly to publish two more (whether any measurement failed, and whether the headline was withheld) so this check can apply `gate.py`'s full definition — *green is no failed measurement and a headline that actually judged*. That is a change for the benchmarks repo; noted in the script rather than silently left.

## Separately — this alone does not make a release installable

`tag-release.yml` creates the GitHub Release with `GH_TOKEN: ${{ github.token }}`, and GitHub does not trigger workflows from events raised by the default token. Every publish workflow triggers on `release: [published]`, so none ran for v1.7.1 — the tag and Release exist, the container registry stops at 1.7.0.

Not fixed here. It needs a PAT with `workflow` scope, or the publish jobs folded into `tag-release.yml` directly.

### DIFF
--- a/scripts/check_release_gate.py
+++ b/scripts/check_release_gate.py
@@ -15,6 +15,35 @@ source-code change.
 not "it is fine" -- treating those as the same thing is the failure the whole
 harness is built to avoid, and a release is the last place to start.
 
+**A green CH-REGRESS run reports `partial`, not `pass`**, so this accepts both.
+The envelope's status is derived pessimistically (benchmarks repo,
+`harness/envelope.py`):
+
+    error      if any measurement errored
+    fail       if any measurement failed        <- a real regression lands here
+    partial    if any measurement is partial or unmeasured
+    pass       otherwise
+
+CH-REGRESS marks each of its 62 per-query timings `partial`, because a timing
+has no pass/fail target of its own -- so the run-level status is `partial`
+however green the suite is, and `pass` is unreachable. Requiring it blocked
+every release from 2026-08-28 (when this check landed) until this fix: twelve
+consecutive verdicts, none of them `pass`, and no tag cut in that window.
+
+The benchmarks repo's own `harness/gate.py` documents the trap it fell into
+here: *"testing for `"pass"` would be a condition that can never hold however
+green the suite is ... It is an easy shape to write and an invisible one to
+read, because 'not met' looks like work remaining rather than like a broken
+test."*
+
+Accepting `partial` does **not** weaken the gate against regressions: by the
+rollup above, a slower query produces `fail`, which is still refused.
+
+It does leave one narrower gap. `partial` also covers "a measurement declined
+to judge", and the verdict file carries no per-measurement detail to tell that
+apart from the ordinary timing case. Closing it needs the nightly to publish
+that detail -- see the note at the bottom of this file.
+
     python3 scripts/check_release_gate.py RELEASE-GATE.json [--max-age-days 3]
 """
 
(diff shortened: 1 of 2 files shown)


Kalyanpulikanti-01:
Closing in favour of fixing the reported status rather than relaxing what the gate accepts.

This PR made `check_release_gate.py` accept `partial` alongside `pass`. That unblocks the tag, but it widens what the gate lets through: `partial` also covers "a headline measurement declined to measure", which is exactly the case a release gate should stop. The right fix is for a green CH-REGRESS run to *report* `pass`, not for the gate to stop asking for it.

That is being done in the harness instead, where the status is produced:

- The envelope's rollup is deliberately pessimistic, and correctly so — it is the internal record, and a timing with no target of its own is genuinely `partial`.
- The *published verdict* is a different thing: it answers "was this run green by `harness/gate.py`'s definition?" It should say `pass` when no measurement failed and no headline measurement was withheld.

So `publish_verdict.py` now derives the published status from `gate.py`'s own definition instead of forwarding the internal rollup. The gate keeps requiring `pass`, unchanged, and a real regression still lands as `fail`.

The analysis in this PR stands and is worth keeping on the record:

- `check_release_gate.py` has required a status CH-REGRESS structurally cannot emit since it landed in `4746793` (2026-08-28).
- Every verdict published since 2026-08-31 has been `partial` or `fail` — **never once `pass`**.
- v1.7.1 (2026-08-27) is the last installable release precisely because it predates the check by one day. The gate has never opened, so this is not a regression in anything; it is a condition that has never been satisfiable.

No change is needed in this repo.

PR #1209, opened 2026-09-14
A named-zone date-time plus a duration re-resolves its offset (#824)

### LINKED ISSUE #824: datetime + duration keeps the old offset instead of re-resolving the named zone
## Symptom

Adding a calendar day to a zoned date-time produces a value whose offset contradicts its own zone name:

```
start         = 2017-10-29T00:00+02:00[Europe/Stockholm]
start + P1D   = 2017-10-30T00:00+02:00[Europe/Stockholm]
```

Stockholm is **+01:00** on 30 October — the fall-back happens at 03:00 on the 29th. There is no such moment as `2017-10-30T00:00+02:00` in that zone, so the value is internally inconsistent: read it back by zone name and you get one instant, read it by offset and you get another, one hour apart.

## Cause

`shift_temporal` carries the stored `offset_seconds` through unchanged. A named zone has no single offset — that was the whole design point of #767 — so the offset has to be **re-resolved from the zone at the new local time**, exactly as `resolve_offset` already does at construction. Only the construction path does it.

## Why it is not visible

The rendered string looks perfectly well-formed, and it is right for every date-time that does not cross a DST boundary — which is almost all of them, and all of the ones a hand-written test would reach for. It goes wrong twice a year per zone.

## Where it came from

Found while writing the tests for #821, not from a failing TCK scenario. I do not believe the TCK covers it: the daylight-saving rows all use `duration.inSeconds` on the *difference* between two temporals rather than on the sum of a temporal and a duration.

Repro is the probe in #821's PR description.

sandeepkunkunuru:
Closes #824.

## Wrong answer on main

`2017-10-29T00:00+02:00[Europe/Stockholm] + P1D` gives `2017-10-30T00:00+02:00`.

Stockholm is +01:00 on 30 October. So the value names one instant by its zone and another by its offset, an hour apart.

## Change

For a date-time in a **named** zone, `shift_temporal` now follows java.time's split, which Neo4j follows too:
- months and days move the local date-time, and the offset is re-resolved from the zone there;
- hours and below move the instant exactly;
- the final offset is the zone's at that instant.

Fixed offsets are unchanged.

## Evidence

`tests/zoned_datetime_arithmetic.rs`: 6/6, read back through Cypher's own accessors. On main, 4 of them fail:

| case | main | now |
|---|---|---|
| `+ P1D` across the October fall-back | 30th 00:00, offset 7200 | 30th 00:00, offset **3600** |
| `+ PT24H` across it | 30th 00:00, offset 7200 | **29th 23:00**, offset 3600 |
| `+ P1D` across the March spring-forward | offset 3600 | offset **7200** |
| `+ P2M` ending after the change | offset 7200 | offset **3600** |
| ordinary day / fixed offset (controls) | unchanged | unchanged |

The 20 other temporal, datetime and duration test binaries pass, and the lib temporal unit tests pass.


sandeepkunkunuru:
Checked against Neo4j 2026.04.0 (local community container, since removed). Every expected value in `tests/zoned_datetime_arithmetic.rs` matches Neo4j's answer:

| start (Europe/Stockholm) | + | Neo4j `[day, hour, minute, offsetSeconds]` | this PR |
|---|---|---|---|
| 2017-10-29 00:00 | `{days: 1}` | 30, 0, 0, 3600 | same |
| 2017-10-29 00:00 | `{hours: 24}` | 29, 23, 0, 3600 | same |
| 2017-03-25 12:00 | `{days: 1}` | 26, 12, 0, 7200 | same |
| 2017-09-29 00:00 | `{months: 1}` | 29, 0, 0, 7200 | same |
| 2017-09-29 00:00 | `{months: 2}` | 29, 0, 0, 3600 | same |
| 2017-10-20 00:00 | `{days: 1}` (control) | 21, 0, 0, 7200 | same |
| `2017-10-29T00:00+02:00` (fixed) | `{days: 1}` | 30, 0, 0, 7200 | same |

sandeepkunkunuru:
Closing and reopening to resync: GitHub still reports head 06eddfb while the branch is at 275c4c4 (main merged in), so CI never ran on the current head.

Issue #1211, opened 2026-09-14
EXISTS { } cannot hold WITH or aggregation: TCK ExistentialSubquery2 [2] is a parse error


sandeepkunkunuru:
## What fails

TCK \`ExistentialSubquery2\` [2], "Full existential subquery with aggregation":

```cypher
MATCH (n) WHERE exists {
  MATCH (n)-->(m)
  WITH n, count(*) AS numConnections
  WHERE numConnections = 3
  RETURN true
}
RETURN n
```

Expected: the one node `(:A {prop:1})`. We return an error. It is 1 of the 3 TCK scenarios that still error. The TCK run on `5c3acd5` evaluated 3,763: 3,760 pass, 0 wrong results, 3 errored.

## Why

- **Grammar:** `src/query/cypher.pest:372` is `exists_subquery = { EXISTS "{" MATCH? pattern where_clause? (RETURN distinct? return_items)? "}" }`. There is no `WITH` and no second clause, so the body above does not parse.
- **AST:** `Expression::ExistsSubquery { pattern, where_clause, bare_pattern }` holds one pattern and one predicate. There is nowhere to put a pipeline.
- **Execution:** 7 evaluation sites in `operator.rs` each handle pattern + WHERE. The only path that runs a full inner query is `execute_call_subquery` (`executor/mod.rs:290`), and that handles the **leading, non-correlated** `CALL { }` only. A full `EXISTS { }` is correlated: the inner `MATCH (n)-->(m)` must use the outer row's `n`.

## Scope

- Parse a full query body inside `EXISTS { }`, and keep today's pattern form as the fast path.
- Run it per outer row with the outer bindings in scope. The answer is true if it yields at least one row.
- The same correlated-execution piece would serve `COUNT { }` and a correlated `CALL { }`. Build it once.

Filed from the night-of-09-14 TCK sweep. Not started.

Issue #1009, opened 2026-08-30
Extended-year date literals need Date widened from i32 to i64 (storage-format change)


sandeepkunkunuru:
```cypher
RETURN duration.between(date('-999999999-01-01'), date('+999999999-12-31'))
```

> `RuntimeError("invalid date: -999999999-01-01")`

openCypher expects `P1999999998Y11M30D`. TCK: `Temporal10[9]` and `Temporal10[10]`.

## This is not a parser fix

Two separate limits, and the second is the real one.

**`chrono::NaiveDate` caps at year ±262,143**, so `parse_iso_date` — which routes everything through `NaiveDate::from_ymd_opt` — cannot construct these. That part could be worked around with proleptic Gregorian day-number arithmetic (Hinnant's `days_from_civil`), independent of chrono.

**But `PropertyValue::Date(i32)` cannot hold the result:**

| | |
|---|---|
| days in 999,999,999 years | 3.652 × 10¹¹ |
| `i32` max | 2.147 × 10⁹ |
| | **170× too large** |

`i32` days spans about **±5.88 million years**. The type's own doc comment says so, and reasons that it is "far past what `chrono::NaiveDate` accepts, so the representation is not the limit". That was true against chrono; it is not true against openCypher, which asks for ±999,999,999.

## What closing it would cost

Widening `Date` to `i64`. That reaches:

- `PropertyValue` and every match on it
- the snapshot format and its backward compatibility
- columnar storage
- `cypher_order` and the derived `Ord` backing the B-tree index
- the SDKs and the wire representation

A storage-format change, in other words, with snapshot-compatibility implications.

## Recommendation: not now

Both scenarios are **errors, not wrong answers** — the engine refuses rather than answering incorrectly, so H1 gate condition 5 is unaffected (it is met at zero). The value is 2 TCK scenarios out of 3,763; the risk is the storage format.

Better done deliberately, alongside any other `PropertyValue` width work, than as a conformance chase. Filing the arithmetic here so the decision is informed rather than rediscovered.

If it *is* taken up, the parser work (chrono-independent civil-date arithmetic) and the width change are separable and the parser half is the easy one.

PR #1222, opened 2026-09-14
A batch lookup probes the index once per row (#1219)

### LINKED ISSUE #1219: Batch lookup UNWIND $rows AS r MATCH (n:N {id: r.id}) should seek the index per row, not scan
## Where this starts

PR #1218 makes `MATCH (n:N {id: r.id})` legal. It rewrites the non-literal property into `WHERE n.id = r.id`. Before that PR the query was an error (TCK Unwind1 [6]).

The answer is correct. The execution is a **correlated filter**: for each unwound row, the plan scans `:N` and keeps the nodes whose `id` equals `r.id`. With an index on `:N(id)`, each row could be a single index seek instead.

## Why it matters

`UNWIND $rows AS r MATCH (n:N {id: r.id}) ... SET/MERGE/CREATE` is how drivers and ETL jobs do batch writes against existing nodes. For B rows and |N| nodes:

| | per batch |
|---|---|
| now (after #1218) | B × \|N\| property comparisons |
| with a per-row index seek | B index probes |

## Not measured yet

Cost claims here are derived from the plan shape, not timed. Before building anything:
- PROFILE the pattern on a real graph with an index;
- time it at, say, B = 10,000 and |N| = 1M, on a quiet host.

## Shape of the fix

This is an index nested-loop join. When a WHERE conjunct is `var.key = expr`, `var` is scanned by label, `key` is indexed, and `expr` reads only variables bound upstream, plan `var` as a per-row index lookup of `expr`'s value.

`IndexScanOperator` exists for literal values; this is its correlated form.

sandeepkunkunuru:
Closes #1219.

## The problem, measured

`UNWIND $rows AS r MATCH (n:N) WHERE n.id = r.id` never used the index on `:N(id)`. The equality spans the MATCH and the UNWIND, so it was applied as a filter on top of a full scan of the label, once for every row. Two plan shapes produce this:
- a leading UNWIND is placed over the match plan;
- a later MATCH is joined to what came before by a cartesian product.

This is how drivers and ETL jobs write to existing nodes in batches.

## Change

- **`CorrelatedIndexLookupOperator`:**
  - for each input row, it evaluates the key and probes the `label.property` index;
  - it emits the row once per live node found;
  - a null key matches nothing, as `n.id = null` does;
  - keys match as the index stores them, exactly as the literal `IndexScanOperator` already matches them.
- **Where the planner uses it:** after a leading UNWIND, and at the two join sites (before and after a WITH).
- **The shape it requires:** one named node with exactly one label, no relationships, no path variable and no inline properties, pinned by an equality to an expression over variables already bound, on an indexed property.
- **What stays:** the node's own WHERE is applied on top. **Every other shape is planned as before.**

## Evidence

Release build on vm-1: 200,000 `:N {id}` nodes with `CREATE INDEX ON :N(id)`, 3 repetitions.

| query | main | this PR |
|---|---|---|
| `UNWIND range(1, 1000) AS i MATCH (n:N) WHERE n.id = i RETURN count(n)` | **62 – 64 s** | **0.6 – 1.8 ms** |
| `UNWIND range(1, 100000) AS i ...` | not run (a full scan per row) | **91 – 95 ms** |
| 1,000 separate `MATCH (n:N {id: <literal>})` (reference) | 10.2 ms | 10.6 – 10.9 ms |

Every run returns the full match count.

- **Answers:** `tests/correlated_index_lookup.rs` compares every answer with the scan's, with and without an index. It covers duplicates, missing and null keys, operands either way round, a filter on the node, a string key, a deleted node, and a lookup after a WITH.
- **Plan:** the same file checks that EXPLAIN shows the new operator only when an index exists, and that a pattern with a relationship keeps its plan. On main the plan check fails; 8/8 pass now.
- **Other tests:** lib 2209/2209. The 45 unwind, match, where, index, join and optional test binaries pass.
- **TCK:** 3,842 of 3,847 evaluated pass. Unwind1 [6] and [14] are fixed by #1218 and #1220; #1211 and #1009 ×2 remain.
- **Clippy:** 792 (ceiling 796).

**Not covered here:** a lookup followed by relationships (`MATCH (n:N {id: r.id})-[:R]->(m)`), and multi-label nodes. Both keep today's plan.


sandeepkunkunuru:
The inline form also takes the new path, now that #1218 is on main. On this branch (`a8d9ad9`, main merged in), with 500 `:N` and an index on `:N(id)`:

`UNWIND [3, 7, 999] AS i MATCH (n:N {id: i}) RETURN count(n)` plans as

    Aggregate
    +- Filter (n.id = i)
       +- CorrelatedIndexLookup (n:N.id = <per row>)

and returns 2, which is correct: 999 does not exist.

#1218 rewrites `{id: i}` into the late-bound predicate, and the lookup finds it there. The Filter above is that same equality checked again on each output row; that costs one evaluation per result and cannot change an answer.

PR #1225, opened 2026-09-14
A full ORDER BY sorts row indices, not (key, record) pairs (IC9 1.12x regression)

sandeepkunkunuru:
## The regression, attributed

CH-REGRESS on main (`09b12de`) was run as a same-host A/B on Vultr against the baseline's commit `0eb6a55`. It failed on one query:

> **IC9: 598 → 668 ms (1.12×)**, confirmed at 1.11× on a second run.

Two bisects on quiet Vultr boxes found where the time went. Both used `ldbc_benchmark --runs 11 --derive-params 50 --query IC9`.

- **Across main:**
  - `0eb6a55` measures 567 ms, and #1198 (`9cfd31a`) measures 662 ms.
  - All 7 later commits, #1203 through #1222, stay within 647–677 ms.
  - A repeat of `0eb6a55` at the end measures 577 ms, so the box did not drift.
- **Across #1198's own 11 commits:**
  - The first 6 are flat at 553–581 ms.
  - `7bf3518` ("ORDER BY: no allocator call per row for its own bookkeeping", #750) measures **613 ms**.
  - `0d2f46d` (ORDER BY borrows a string key from its column) measures **643 ms**.
  - `5263948` measures 640 ms, no change.
  - A repeat of `0eb6a55` at the end measures 559 ms, so that box did not drift either.

The per-operator profile (`--profile`, vm-1, the three commits back to back) puts all of it in `Sort`. Expand, both filters and the variable-length expand are unchanged across all three.

| commit | Sort self | IC9, uninstrumented |
|---|---:|---:|
| `0226167` (before #1198's sort changes) | 131 ms | 285 ms |
| `7bf3518` | 192 ms | 348 ms |
| `0d2f46d` = main | 200 ms | 348 ms |
| **this PR** | **153 ms** | **290 ms** |

## Why

IC9's plan is `Limit(20)` over `Distinct` over `Project` over `Sort`. The DISTINCT stops the LIMIT from reaching the sort, so `Sort` **fully sorts about 467,000 rows**.
- `sort_by` moved each `(key, record)` pair at every merge step.
- `7bf3518` stored the key inline (`[KeyPart; 2]`) to save an allocation per row. That made each pair about 130 bytes instead of about 48, so the full sort copied far more memory.

A first attempt, a bounded top-k for a small LIMIT, measured no change (Sort 208 ms). That is what showed the LIMIT never reaches this sort. It is not in this PR.

## Change

When no LIMIT reaches `SortOperator`:
- the rows and their keys stay where they are;
- only 4-byte row indices are sorted, compared through the keys;
- the index breaks ties, so the order is stable and deterministic, as `sort_by`'s was.

The top-k path under a LIMIT is unchanged.

Sort self time is still about 20 ms above its pre-#1198 131 ms, most likely `0d2f46d`'s per-row string-borrow attempt on a non-string key. The query as a whole is within noise of its pre-#1198 time.

## Evidence

- **IC9 profile (vm-1):** 348 → **290 ms**, against 285 ms before #1198. Sort self time 200 → **153 ms**.
- **CH-REGRESS same-host A/B on Vultr** against this commit (`edd9f9f`, arm A `0eb6a55`): running at the time of opening; the verdict is added below when it finishes.
- `tests/sort_top_k.rs` checks that ORDER BY … LIMIT k equals the first k rows of the full sort. It covers both directions, SKIP, nulls at either end, two keys, ties at the cutoff, and a LIMIT larger than the input. It passes before and after.
- `order_by_key_errors`: 5/5. Lib: 2209/2209.
- The 37 order, sort, limit, skip, return, WITH, distinct and aggregation test binaries pass.
- TCK: 3,844 of 3,847 evaluated pass, 0 wrong.
- Clippy: 792 (ceiling 796).


sandeepkunkunuru:
**CH-REGRESS confirmation, measured.** This ran as a same-host A/B on Vultr `voc-m-16c-128gb`: the baseline's commit `0eb6a55` (arm A, seeded on the box) against this PR's commit `edd9f9f` (arm B).

- **Verdict: `pass`.** All 62 queries are within 1.1× and +5 ms, confirmed over 2 separate runs.
- Preconditions: the host was quiet (median 0 other runnable threads over 743 samples), and calibration was 43.0 ms at both arms.
- **IC9: 582 → 563 ms (0.97×).** The run on main before this PR had 1.12× (598 → 668 ms).
- Breached on the first pass and not on the re-run, so not counted:

  | query | first pass | re-run |
  |---|---:|---:|
  | BI-9 | 1.2× (+87 ms) | 1.05× |
  | IC12 | 1.1× (+7 ms) | 1.06× |
  | IC14 | 1.2× (+12 ms) | 1.00× |

- Improved:

  | query | ratio |
  |---|---:|
  | IC1 | 0.17× |
  | IC6 | 0.53× |
  | BI-12 | 0.63× |
  | IC11 | 0.64× |
  | BI-4 | 0.84× |
  | BI-11 | 0.85× |
  | BI-16 | 0.88× |

The envelopes are kept locally and are **not published** to release-gate. Whether a Vultr same-host A/B counts for the v1.8.0 gate is the owner's decision. The box was destroyed after the run, and 0 instances remain.

Issue #1200, opened 2026-09-14
MVCC as an undo log beside the columns: real transactions without storing every property twice (#1188, #545)


sandeepkunkunuru:
**Design for review. No code until it is agreed.** This unblocks #1188 and #545.

## Decision (owner, 2026-09-14)

- **Keep transactions as a product requirement.** They are needed for:

  | row | what it asks for |
  |---|---|
  | LANG-07 | BEGIN / COMMIT / ROLLBACK with documented isolation |
  | REL-01 | a published isolation-anomaly table |
  | REL-02 | Jepsen-style consistency testing |
  | REL-05 | point-in-time recovery |
  | PERF-16 | ≥200K node writes/s with MVCC on, p99 < 10 ms |
  | PERF-17 | 64 concurrent clients |
  | INT-03 | Bolt, whose sessions are built on explicit transactions |

- **Drop the per-entity full copies.** History moves into a per-version log kept beside the columns.

## What exists today

From reading the code at `9cfd31a`. Nothing here was measured unless the text says so.

**Node properties.** `set_node_property` always overwrites the column. It then writes the row copy in `Node.properties`. If the node's newest version is older than `current_version`, it first clones the **whole `Node`**, labels and full `PropertyMap` included, and pushes it onto `nodes: Vec<Vec<Node>>`. `get_node_at_version(id, V)` returns the newest version ≤ V.

**Edge properties.** The row map, plus `edge_version_log`. Since #602 the log is lazy: no entry at the first version, and an entry is sealed only when a later version changes the edge.

**Transactions:**
- `begin_transaction` records `start_version` and empty write sets.
- `txn_write_node` and `txn_write_edge` only add an id to the write set. **The write itself goes straight to the live store.**
- `commit_transaction` checks that no write-set entity was committed after `start_version` (first-committer-wins), then bumps `current_version`.
- `abort_transaction` only marks the transaction Aborted. The code says so: *"actual rollback of in-place mutations requires version-aware cleanup."*
- ReadCommitted reads at `current_version`; SnapshotIsolation reads at `start_version`.

**So the current API does not provide isolation:**
- Uncommitted writes are visible to every reader: dirty reads.
- A snapshot taken at version V sees writes made later at that same V, because writes do not advance the version.
- ROLLBACK does not exist.

[...]

Issue #1188, opened 2026-09-10
[PERF-10] Stop writing the node row copy: 2,109 MB on SF1, blocked on MVCC property history


sandeepkunkunuru:
## Stop writing the node row copy — attempted, measured, and what still blocks it

Follows #545 and #1123. Node properties are stored twice. `set_node_property` and `create_node_with_properties` write every value to the column store **and** to the row `Node.properties`. On LDBC SF1 at `717cde3`:

```
RSS bytes/edge (PERF-10)          531.8
heap bytes/edge                   456.5
node properties (row)     2,108,615,887   26.8% of live heap  ≈ 122 B/edge
node properties (columnar)  944,598,850   12.0%
```

The change was built — clear the row key on every store write, create nodes with an empty row, make the column win on a clash — and run against the full suite. **40 tests failed across 16 binaries.** Several were wrong query answers, not tests pinning the old layout. It was pulled from PR #1186, which ships the readers it exposed instead.

## Resolved by #1186

- **Eight row-only readers** via `Node::get_property` — #1187, fixed. Five were already wrong on restored graphs.
- **Four readers** via `node.properties` — `node_matches` (MERGE, #1185), `schema_summary`, vector discovery and export `json_of`.

## Still blocking

1. **MVCC property history.** Each node version carries its own `PropertyMap`; the column holds only the current value. The failing tests `test_mvcc_node_versioning`, `test_mvcc_cow_versioning`, `test_snapshot_isolation_reads_start_version`, `test_read_committed_sees_latest`, `test_gc_node_versions`, and `mvcc_test.rs` ×2 read historical values through the transaction API. **Nothing outside tests calls `begin_transaction` or `commit_transaction`**, so `current_version` never advances in production and no node ever has a second version. But the API is public and documented. Deleting the row copy removes a working feature, even if nobody uses it. That's a product decision: either keep the row for versioned nodes only, or move history into a per-version column log. It isn't a refactor.
2. **Test fix-ups — 43 known.** 34 value assertions would fail. 9 absence assertions would pass *vacuously*, since an empty row lacks every key, so REMOVE could break without any of them noticing. A rewrite script exists; one `let x = …get(k)` site needs a hand fix for a temporary's lifetime.
[...]

PR #1261, opened 2026-09-15
MVCC step 3: a node property is stored once, in the column (#1188)

sandeepkunkunuru:
Step 3 of #1200, and #1188 itself. A node property is stored once, in the column. Stacked on #1260 (step 2); this PR's base is that branch.

## Why

`set_node_property` and `create_node_with_properties` wrote every value twice: once to the column store and once to the node's row `PropertyMap`. On LDBC SF1 at `717cde3` the row copy was **2,108,615,887 B**, 26.8% of the live heap, about 122 B/edge (#1188).

It was blocked on MVCC history, because each node version carried its own map. Step 2 moved history into an undo log that reads the column first, so the row no longer carries anything history needs.

## What changes

- **`create_node_with_properties` leaves the row empty.** The map moves into the index event rather than being cloned (#1132 already built the event from the properties as given).
- **`set_node_property` no longer writes the row.**
  - It reads the old value column-first, once. That value goes to both the undo entry and the index event.
  - This matters for the index: the old value used to come from the row's `insert`. With an empty row that returns `None`, so the index would keep the stale entry, and a lookup by the replaced value would still find the node. `an_index_drops_the_value_a_write_replaced` pins this.
  - It removes the key from the row. A value put in the row directly through `get_node_mut` is superseded by a store write, rather than shadowing the column.
- **`node_materialized` and `node_properties_full` let the column win** and keep row-only keys, the way `node_properties_merged` already did. Before, `node_materialized` merged only when the column had a key the row lacked, so a superseded row value won.
- **Unchanged:** writes through `get_node_mut().set_property(..)` still go to the row only, and every reader falls back to the row. `main.rs`'s demo, `agent/executor.rs` and `uc_problems.rs` keep working and keep their row values. Moving them onto `set_node_property` would shrink the row further, and can follow separately.

## Tests

- **New: `tests/node_row_copy.rs`, 5 tests.**
  - No row copy after a create with properties.
  - No row copy after a property write.
  - A store write supersedes a row-only value in the merged map.
  - A row-only value is still read, and both views agree.
  - An index drops the value a write replaced.
- **Rewritten:** 57 assertions that read the row, now through `store.node_property(id, key)` (column first, then row). There are 38 in `query/executor` tests, 10 in `graph::store` tests, and 9 across 9 files in `tests/`: arrow export, auto-embed, the count shortcut, `load_csv`, `mvcc_test`, optimization, `result_cache` ×2, vector search, and the pinned var-length walk.
- **Triage against #1188's earlier attempt.**
  - That attempt failed 40 tests across 16 binaries. On this branch the same binaries give **44 failures before the rewrite, and every one was a test reading the row**.
  - The two candidates for a real wrong answer turned out to be test helpers:
    - the pinned var-length operator "finding nothing" was the test's `walk` reading names off `get_node(..).properties`;
    - export's "3 embeddings, not 2" was the fixture filtering on the row.
  - The other binaries from that list (anchor coverage, projection semantics, expand-target, GQL path modes, record capacity, v0.6 features, var-length enumeration) pass unchanged. #1186's reader fixes and the column-wins merge cover them.

## Results

| run (vm-1, debug) | result |
|---|---|
| `cargo test --lib` plus 24 binaries: every one that failed in #1188's attempt, every file that read the row, `node_row_copy` and `mvcc_isolation_anomalies` | 2,467 passed, 0 failed |

The full workspace suite is left to CI.

**SF1 memory and `RETURN n` latency before and after.** These come from a GitHub runner (the AWS account is blocked), and I'll post them as a comment here. `benches/memory_footprint.rs` gains `--return-n` for that: it times execute plus materialising each returned node. SF10 needs a 128 GB host, so it waits for AWS or an approved Vultr box.

Refs #1188, #545, #1200.


sandeepkunkunuru:
**SF1 measurement, before and after (#1188).** Run 34988374848 on a GitHub runner (the AWS account is blocked), two interleaved rounds, both trees built from the same `benches/memory_footprint.rs`. base = step 2 (`cfabdc3`), fix = step 3 (`d2938f1`). LDBC SF1: 3,181,724 nodes, 17,256,038 edges.

## Memory

| | base | fix | change |
|---|---:|---:|---:|
| live heap | 7,688,341,657 B | 5,273,854,663 B | **−31.4%** |
| RSS | 8,976,936,960 B | 5,888,503,808 B | **−34.4%** |
| heap B/edge | 445.5 | 305.6 | −31.4% |
| **RSS B/edge (PERF-10)** | **520.2** | **341.2** | **−34.4%** |
| node row properties | 2,108,615,887 B (27.4% of heap) | 2,597,673 B (0.0%) | −99.9% |
| node side (versions + row + labels) | 2,849,910,242 B | 743,892,028 B | −73.9% |

Round 2 agrees to the byte on heap and within 0.1% on RSS (RSS B/edge 520.0 → 341.6).

The 2.6 MB of row properties left are the values written through `get_node_mut`, which this PR leaves alone.

**PERF-10's target is 128 B/edge.** At 341 B/edge it is still not met; this removes the largest single contributor.

## `RETURN n` latency (execute + materialise, p50 over 50 runs)

| query | base r1 / r2 | fix r1 / r2 |
|---|---|---|
| `RETURN n`, 1,000 Person | 4.58 / 4.71 ms | 4.86 / 4.94 ms |
| `RETURN n`, 10,000 Post | 55.86 / 58.82 ms | 62.71 / 66.06 ms |
| `RETURN n.firstName` (control) | 0.51 / 0.52 ms | 0.38 / 0.36 ms |

Returning whole nodes costs **6–12% more**: the map is rebuilt from the columns instead of cloned from the row. Returning one property is **~30% faster**, which is the same change seen from the other side — there is no row map to fall back to and no second copy to keep warm.

That is the trade #1188 predicted: 2.1 GB against single-digit percent on `RETURN n`.

**SF10** needs a 128 GB host, so it waits for AWS or an approved Vultr box.

sandeepkunkunuru:
Closed automatically when #1260 was merged with `--delete-branch`, which deleted this PR's base branch; GitHub will not reopen it.

The work is intact and now lands through **#1264**, which carries steps 3, 4, 5a and 5b against main. The SF1 before/after measurement is in the comment above and is quoted there.

PR #1262, opened 2026-09-15
MVCC step 4: transactions buffer their writes; ROLLBACK drops them (#1200)

sandeepkunkunuru:
Step 4 of #1200: transactions buffer their writes, and ROLLBACK drops them. Stacked on #1261 (step 3), which is this PR's base.

## Why

A transaction's writes went straight to the live store:
- every reader saw them before commit (dirty reads);
- `abort_transaction` only flagged the transaction, so there was no rollback;
- a snapshot at version V saw writes made at V, because writes did not advance the version.

Step 1's isolation-anomaly tests (#1206) pinned four such failures.

## What changes

- **`txn_set_node_property(txn, node, key, value)`** buffers a write in the transaction (`Transaction.node_writes`). As on the store, `Null` removes the key.
- **`txn_create_node(txn, labels)`** reserves an id and records the node (`Transaction.created_nodes`). The node enters the store at commit under the same id.
- **`get_node_for_txn`** reads the transaction's own writes and created nodes over its snapshot. No other reader sees them.
- **`commit_transaction`** checks conflicts first-committer-wins, as before. It then bumps the version and applies the buffer through the store's writers. So the undo log (step 2) records the writes at the new version, and created nodes are born at it. The apply is all or nothing: on the first error it puts back what it applied and removes the nodes it created.
- **`abort_transaction`** drops the buffer and returns reserved ids to the free list.
- **`create_node_with_labels`** is split into `allocate_node_id` and `insert_node_with_labels`, so the id a transaction reserves is the id its node keeps.

The decisions recorded on #1200 (2026-09-15) hold:
- conflicts are checked per entity;
- snapshot isolation is the default, with write skew allowed;
- a transaction reads its own writes through an overlay.

## Tests: `tests/mvcc_isolation_anomalies.rs`, 10 of 10

- **The four step-1 tests that were ignored now pass:** dirty read, rollback, a snapshot after a later commit, and phantom by creation. The creation test uses `txn_create_node`; the old test created the node outside any transaction.
- **Unchanged:** lost update and write skew.
- **New:**
  - a transaction reads its own writes;
  - a removal is buffered too;
  - an aborted create leaves no node and returns its id;
  - a commit refused by a unique constraint changes nothing. The transaction also wrote another node and created one; both are rolled back.

## Results (vm-1, debug)

| run | result |
|---|---|
| `--lib`, `-p samyama-sdk`, `node_row_copy`, `mvcc_test`, `mvcc_isolation_anomalies` | 2,277 passed, 0 failed |
| clippy on lines this stack added | 0 new lints (local 1.96) |

## Not covered here

- **Relationship writes inside a transaction.** Step 5 moves relationships onto the same undo log.
- **Deletes inside a transaction.** They still go straight to the store.
- **BEGIN, COMMIT and ROLLBACK over HTTP and RESP.** That is step 6.


sandeepkunkunuru:
Superseded by #1264, which now carries steps 3, 4, 5a and 5b as four commits against main.

This branch's work is unchanged and included there; its CI run on this head was green. The stack broke when step 2 (#1260) was merged with `--delete-branch`: that deleted #1261's base branch, GitHub closed #1261 and refuses to reopen it, and main now holds step 2 as one squashed commit, so each remaining branch conflicts against it. Rebuilding the stack link by link would re-conflict after every merge into main.

Closing as superseded rather than merging.

PR #1263, opened 2026-09-15
MVCC step 5a: relationship history as the same undo log as nodes (#1200)

sandeepkunkunuru:
Step 5a of #1200: relationship history uses the same undo log as node history (step 2). Stacked on #1262 (step 4), which is this PR's base.

## What changes

- **`edge_version_log` is gone.** It kept a copy of an edge's whole property map for each version it was written at; since #602 that copy was taken lazily, at the first write at a later version.
- **It is replaced by `edge_history`:**
  - a birth version, recorded only when the edge is created after version 1;
  - `EdgeUndo::Property { at, key, old }` entries: one per key, for each newer version that overwrote the key.
- **Reading at V** takes the live map and undoes the entries newer than V, newest first, as for nodes.
- **Writers:** `set_edge_property`, `set_edge_property_sparse` and `set_edge_properties_sparse` record the value they replace before writing. `remove_edge_property` records one key; it no longer seals the whole map.
- **`get_edge_properties_mut`** is the one path that records the whole map (`EdgeUndo::Map`), because it does not say which key the caller will change.
- **GC, `delete_edge` and `clear`** follow the node side. A reused edge id inherits no history.
- **A store whose version never advances records nothing**, which is the #602 property.

## Tests

- **`tests/edge_version_history.rs`** passes unchanged. It has 13 cases covering every setter, the first-version rule and edges created after a commit.
- **`test_snapshot_isolation_edge_read` and `test_gc_edge_version_log`** used to push log entries by hand. They now write at real versions; the GC pruned count is unchanged.
- **New:**
  - changing one key out of 10 records one `Property` entry, not a copy of the map;
  - no history is kept when the version never advances;
  - a key added through the mutable map is absent at an earlier version;
  - a reused edge id does not inherit the deleted edge's history.

## Results (vm-1, debug)

| run | result |
|---|---|
| `--lib`, `-p samyama-sdk`, `edge_version_history`, `mvcc_test`, `mvcc_isolation_anomalies`, `node_row_copy` | 2,294 passed, 0 failed |
| clippy on lines this stack added | 0 new; the local total fell 769 → 767 |

I'll post the mutation check (undo push removed from `log_edge_write`) as a comment.

**Next, step 5b:** stop writing the relationship row copy, the other half of #545.


sandeepkunkunuru:
**Mutation check.** I removed the undo push from `log_edge_write`. With nothing recording what an edge write replaces, 12 tests fail:

- 9 of the 13 in `tests/edge_version_history.rs`: every setter, the first-version rule, and edges created after a commit;
- 3 in `graph::store`: `test_edge_history_is_one_entry_per_changed_key_not_a_map_copy`, `test_gc_edge_version_log` and `test_snapshot_isolation_edge_read`.

The 4 that still pass assert behaviour at the current version or on a missing edge, which history does not affect. The file was restored with `git checkout`.

sandeepkunkunuru:
Superseded by #1264, which now carries steps 3, 4, 5a and 5b as four commits against main.

This branch's work is unchanged and included there; its CI run on this head was green. The stack broke when step 2 (#1260) was merged with `--delete-branch`: that deleted #1261's base branch, GitHub closed #1261 and refuses to reopen it, and main now holds step 2 as one squashed commit, so each remaining branch conflicts against it. Rebuilding the stack link by link would re-conflict after every merge into main.

Closing as superseded rather than merging.

PR #1265, opened 2026-09-16
MVCC step 6a: session transactions in the store (BEGIN, COMMIT, ROLLBACK) (#1200)

sandeepkunkunuru:
Step 6a of #1200: session transactions in the store. Against main: #1264 (steps 3-5b) has landed as `dd50c40`, and this branch now carries main plus the step-6a commit.

## The approach

The owner picked **session lock plus undo** over an executor overlay (#1200, 2026-09-16): a transaction's statements run directly against the store while the caller holds the writer's lock, and everything they change is recorded so ROLLBACK can put it back. There are no dirty reads because no other reader can take the lock while the transaction is open. The cost is that an open transaction blocks readers, which step 6b bounds with a timeout.

## What this PR adds

- **`begin_session_transaction`** advances the version and opens a journal. Writes then land at the new version, so the undo logs of steps 2 and 5a already record what each replaced and which entities were born.
- **`commit_session_transaction`** keeps everything and closes the journal.
- **`rollback_session_transaction`**, in order:
  1. restores what the transaction deleted, nodes before the relationships that need them, including the catalog counts `delete_node` decremented and the ids it freed;
  2. undoes every write recorded at its version, newest first, **through the ordinary setters**, so the property index, label index and catalog stay in step;
  3. removes what the transaction created;
  4. drops the history it wrote and puts the version back.
- **Deletion capture.** A deletion is the one change the undo logs cannot reconstruct, so `delete_node` and `delete_edge` capture the entity **as of the version before the transaction**. Capturing it as it stands would restore a node that was written and then deleted in the same transaction with the value that transaction wrote, because deleting drops the history that would have undone it. `test_rollback_restores_a_node_written_then_deleted_in_one_transaction` pins this.

## Tests (6 new)

- rollback of a property write and of a label;
- rollback removes what the transaction created, and leaves no history;
- rollback restores a deleted node, its properties, its label index entry and its relationships;
- rollback of a write-then-delete restores the pre-transaction value;
- commit keeps the writes and the version, and the committed write reads as history at the earlier version;
- one session transaction at a time; committing twice and rolling back with none open both fail.

## Results (vm-1, debug)

`--lib`, isolation anomalies, MVCC, node and relationship row-copy suites: **2267 passed, 0 failed**

**Mutation check:** with the deleted-node capture removed, the rollback restore tests fail (output in the run log); the file was restored with a checked copy.

## Next: step 6b

BEGIN/COMMIT/ROLLBACK over HTTP and RESP (LANG-07), holding the write lock per RESP connection or HTTP session with a timeout, plus the OpenAPI entries CH-API checks and the published isolation table (REL-01).


sandeepkunkunuru:
**Mutation check.** I removed the deleted-node capture in `delete_node` (the `txn.deleted_nodes.push`), so a rollback has nothing to restore a deleted node from:

```
test test_rollback_undoes_a_property_write ... ok
test test_rollback_removes_what_the_transaction_created ... ok
test test_rollback_restores_a_deleted_node_with_its_relationships ... FAILED
test test_rollback_restores_a_node_written_then_deleted_in_one_transaction ... FAILED
test result: FAILED. 2 passed; 2 failed
```

Exactly the two restore tests fail; the two that roll back writes and creations still pass, because those paths use the undo log rather than the capture. The file was restored from a checked copy (0 mutant lines left).

Issue #1269, opened 2026-09-16
Embedded (library) users on glibc still carry MVCC step 5b's 1.5-1.7x BI regression: find and fix the load-path fragmentation


sandeepkunkunuru:
Follow-up to ADR-038 (#1267, #1268).

## Problem

ADR-038 installs mimalloc as the global allocator in the **server** and in the benchmark binaries. A library cannot set the allocator for its host process, so everyone who **embeds** the engine stays on glibc: the Python wheel, the Rust SDK (`EmbeddedClient`) and any Rust application depending on `samyama`. On glibc, MVCC step 5b (`97e97ea`, #1264) left BI-9, BI-12 and BI-14 1.5–1.7× slower (runs 35104224715, 35113288768).

## What is known

After an SF1 load, glibc holds **797 K free chunks** on main against **219 K** before 5b (`mallinfo2`, run 35126988660). The profile shows the cost in `_int_malloc`, `malloc_consolidate` and `unlink_chunk`.

Three candidate fixes did not move the chunk count or the queries:
- **borrowed property keys at load**, so no key `String` is allocated and freed per property;
- **`malloc_trim` after load**;
- **`glibc.malloc.mxfast=0`**, which does help some BI queries but makes point reads 1.4–2.0× slower and gets worse under write churn.

## What is not known

**Which load-time allocation pattern leaves the chunks.** They follow the removal of the per-relationship row maps. The plausible reading is that the long-lived maps used to fill holes the column store leaves as it grows (`Vec` or `HashMap` growth in `edge_columns`); it has not been shown.

## Plan

1. **Measure where the chunks come from.** Snapshot `mallinfo2` per load phase (nodes, each edge file, compaction), then bracket the phase where the count jumps.
2. **Test candidates in that phase.**
   - Pre-size the column store from row counts, which the loaders and `.sgsnap` headers know.
   - Rebuild columns into exact-size allocations after bulk load.
   - Allocate column growth in fewer, larger steps.
3. **Acceptance.**
   - glibc free chunks after an SF1 load within 1.5× of pre-5b (219 K).
   - BI-9, BI-12 and BI-14 on glibc within 1.1× of pre-5b, on one runner.
   - No RSS regression at SF1.
4. **Cover snapshot import too.** It is the path every published KG takes, and it churns more per relationship than the LDBC loader.

**Workaround until then:** embedded Linux users can run with `LD_PRELOAD` pointing at mimalloc or jemalloc.

[...]

PR #1270, opened 2026-09-16
mimalloc as the global allocator for the server and benchmark binaries (ADR-038)

### LINKED ISSUE #1267: ADR-038: memory allocator and load-path allocation — MVCC step 5b's BI regression is glibc heap state
## Problem

CH-REGRESS fails on main. On a same-host Vultr A/B on 2026-09-16, `531771f` was slower than `0eb6a55` on five queries, each confirmed on a second run:

| query | slowdown |
|---|---|
| BI-9 | 1.8× |
| BI-14 | 1.7× |
| BI-8 | 1.4× |
| IC14 | 1.3× |
| IC5 | 1.2× |

H1 condition 3 cannot close while this stands.

It bisects to **one commit: MVCC step 5b** (`97e97ea`, #1264), which dropped the relationship row copy. Bisect runs on GitHub runners, each an A/B on one runner:

| run | range | result |
|---|---|---|
| 35099052379 | executor commits, MVCC step 2 | flat |
| 35099052379 | `dd50c40` (steps 3–5b) | BI-12 1.62×, BI-14 1.46×, BI-9 1.43× |
| 35104224715 | steps 3, 4, 5a | flat |
| 35104224715 | **step 5b** | **BI-14 1.68×, BI-12 1.59×, BI-9 1.45×, BI-8 1.27×, BI-4 1.26×** |

## It is glibc heap state, not query code

**Profile** (run 35109413988, BI-14, `perf`, flat self time). The engine's own operators take the same share before and after. Step 5b adds about 20 points inside glibc's allocator. With symbols resolved (run 35113288768):
- `_int_malloc` 29.6%;
- `cfree` 13.2%;
- `malloc_consolidate` 7.7%;
- `_int_free` 4.0%;
- `unlink_chunk` 2.4%.

**Allocator swap** (run 35113288768, same runner, 3 interleaved rounds, median). Under jemalloc (`LD_PRELOAD`) the regression disappears:

| query | glibc before | glibc 5b | ratio | jemalloc before | jemalloc 5b | ratio |
|---|---:|---:|---:|---:|---:|---:|
| BI-14 | 667 ms | 1,025 ms | 1.54× | 495 ms | 507 ms | 1.02× |
| BI-12 | 936 | 1,414 | 1.51× | 661 | 671 | 1.02× |
| BI-9 | 550 | 819 | 1.49× | 352 | 346 | 0.98× |
| BI-8 | 558 | 665 | 1.19× | 526 | 493 | 0.94× |

jemalloc was also faster **before** step 5b: BI-1 1,418→873 ms, BI-2 5,751→4,591 ms, BI-11 1,116→829 ms, BI-14 667→495 ms.

**Likely mechanism (inferred, not yet measured).** The LDBC and FinBench loaders call `set_edge_properties_sparse(edge_id, parse_props(..))`, which builds a `(String, PropertyValue)` list for each relationship.
- Before 5b, the key strings moved into the row map and stayed.
- After 5b, each key string is freed as soon as its value is in a column: tens of millions of small frees during load.

glibc keeps these in fastbins and consolidates them lazily. The cost lands on later allocations, which is to say on queries.

## The decision this needs (ADR-038)

The engine's allocator, and how load paths allocate, are architectural. They change resident memory (the PERF-10 target), add a dependency on every shipped platform, and interact with the library crates.

**Options:**

- **A. Remove the churn.** Load paths write relationship (and node) properties straight into the columns, with no per-entity map, list or key `String`, so no free-list debris. glibc stays.
- **B. Return the debris after load.** Call `malloc_trim` / `malloc_consolidate` at the end of bulk load. glibc stays.
- **C. glibc tunables.** For example `glibc.malloc.mxfast=0` (no fastbins) or trim thresholds, set in the server binary or the Docker image.
- **D. jemalloc as the global allocator** in the **binaries only**: the `samyama` server and `samyama-cli`, via `tikv-jemallocator`.
- **E. mimalloc as the global allocator,** binaries only.
- **Combinations,** for example A together with D.

**Constraints:**
- The Python wheel (`sdk/python`, `cdylib`) and the Rust SDK are libraries. A library must not set `#[global_allocator]` for the host process, so D and E cannot reach embedded users, and those users would still see the regression.
- Shipped targets: Docker `linux/amd64` and `linux/arm64` on `debian:bookworm-slim` (glibc); PyPI manylinux_2_28 x86_64 wheels plus a macOS wheel; npm.
- PERF-10 is an **RSS** target (207 B/edge at SF10 on `531771f`), so an allocator's retained memory counts against it.

## Analysis before the ADR (no fix merged until it is decided)

1. **Mechanism, measured.** Count allocations and frees during load before and after 5b (`benches/memory_footprint.rs` has a counting allocator), and read glibc's free-chunk state after load (`mallinfo2`: `fordblks`, `fsmblks`, `smblks`).
2. **Which real paths are affected.** Check `.sgsnap` import, CSV and JSON import, `LOAD CSV`, Cypher `CREATE`, and the bulk loaders for per-entity map or key churn. If only the benchmark loaders churn, the regression is partly a benchmark artifact, and the ADR must say so.
3. **Options A–E, same runner, SF1.** Measure:
   - all CH-REGRESS queries, p50 over interleaved rounds;
   - load time;
   - live heap and RSS after load.
4. **Long-running server.** A write-churn workload (CREATE, SET, DELETE, snapshot import and re-import) for N minutes per option, tracking RSS drift and latency drift. Fragmentation from churn is not load-only.
5. **SF10 RSS for the finalists** (PERF-10). This needs a 192 GB host; AWS is still blocked, so it needs Vultr approval.
6. **Platforms and build.** Check that the finalists build and run on the Docker targets (amd64 and arm64), the macOS wheel, and CI; note binary size and licence.

**Deliverable:** `docs/ADR/ADR-038-memory-allocator-and-load-path-allocation.md`, with the measurements, the decision, and what embedded (library) users get.

Refs #1200, #1264, #1188, #545. PERF-10, PERF-18, H1 condition 3.

sandeepkunkunuru:
Implements ADR-038 (#1268), closes #1267.

## Why

On glibc, MVCC step 5b (#1264) made BI-9, BI-12 and BI-14 1.5–1.7× slower and failed CH-REGRESS. The cause is glibc's heap after load, not query code.

ADR-038 chose mimalloc:

| | glibc | mimalloc |
|---|---:|---:|
| BI-14 (SF1) | 1.00× | **0.46×** |
| BI-12 (SF1) | 1.00× | **0.47×** |
| BI-9 (SF1) | 1.00× | **0.36×** |
| queries >2% faster (of 41, SF1) | — | **38** |
| BI-14 under 10 min of write churn | drifts upward | flat |
| PERF-10 (SF10) | 207.1 B/edge | **199.3 B/edge** |

## What changes

- **`samyama::allocator`** names the shipped allocator and **never installs it**:
  - `Shipped` / `SHIPPED` are mimalloc under the new **default** feature `mimalloc`;
  - with `--no-default-features` they are `std::alloc::System`;
  - `NAME` gives the allocator's name for output.
- **Who installs it.** The server (`src/main.rs`) and the gated benches: `ldbc_benchmark`, `ldbc_bi_benchmark` and `finbench_benchmark` install `SHIPPED`; `memory_footprint`'s counting allocator wraps it instead of `System`. Each bench prints the allocator.
- **Embedded users are unaffected.** `samyama-sdk` depends on `samyama` with `default-features = false`, so the Python wheel and embedded users keep their host's allocator and never compile mimalloc. `cargo tree` confirms mimalloc is absent from `samyama-sdk` and from `sdk/python`.
- **Guard: `tests/allocator_consistency.rs`.** It checks that:
  - the default build is mimalloc;
  - the server and every gated bench install `SHIPPED`;
  - the library installs nothing;
  - the SDK opts out of the engine's default features.

## Verified

| run (vm-1, debug) | result |
|---|---|
| `--lib` | 2,239 passed |
| `allocator_consistency` | 5 passed |
| **mutant:** install removed from `ldbc_bi_benchmark` | `every_gated_bench_installs_the_shipped_allocator` fails |
| `cargo check` default, `--no-default-features`, `-p samyama-sdk` | all compile |
| clippy on added lines | 0 new lints |
| builds on Linux amd64, Linux arm64, macOS arm64 (ADR run 35127548751) | pass |

A same-runner A/B of main against this branch, across all BI and IC queries and using the real `mimalloc` crate rather than `LD_PRELOAD`, is running; results will be posted here.

## Consequences (ADR-038)

- **Every latency and footprint number moves.** Envelopes should say which allocator produced them.
- **CH-REGRESS baseline.** It was seeded on glibc and must be re-seeded on the fixed host. Until then, a pass after this change does not show the absence of a code regression.
- **Idle RSS** is about +27 MB on Linux.
- **Embedded glibc users** keep the step-5b regression until #1269.


sandeepkunkunuru:
**Same-runner verification** (run 35140540502): main `e2d5604` on glibc against this PR's `da315ef` on mimalloc, with the allocator installed by the benches themselves (no `LD_PRELOAD`). LDBC SF1, every BI and IC query, 3 interleaved rounds, median of per-query medians.

| query | glibc | mimalloc | ratio |
|---|---:|---:|---:|
| BI-9 | 729 ms | 320 ms | **0.44×** |
| BI-12 | 1,219 | 626 | **0.51×** |
| BI-14 | 836 | 448 | **0.54×** |
| IC2 | 4.3 | 2.5 | 0.58× |
| BI-4 | 252 | 159 | 0.63× |
| BI-8 | 613 | 450 | 0.73× |
| BI-2 | 5,122 | 3,739 | 0.73× |
| IC9 | 268 | 199 | 0.74× |
| BI-11 | 1,038 | 810 | 0.78× |
| BI-1 | 1,000 | 790 | 0.79× |
| IC5 | 336 | 269 | 0.80× |
| IC14 | 35.4 | 28.9 | 0.82× |

- **Faster:** 28 of 34 queries by at least 5%.
- **Unchanged:** IC12 1.01×, IC1 1.00×, BI-3 1.00×, IC3 0.99×, IC6 0.97×.
- **Slowest:** BI-19 at 1.09×, 3.3 → 3.6 ms, which is +0.3 ms. That is below CH-REGRESS's regression threshold, which requires both 1.1× and +5 ms.

These agree with the `LD_PRELOAD` measurements in ADR-038.

**This run also found a defect in this PR, now fixed in `2c28009`.** Neither arm printed its `Allocator:` line. My script had inserted the line after the first `eprintln!` in `main`, which in `ldbc_bi_benchmark` sits inside the missing-data-directory branch and in `ldbc_benchmark` inside the unknown-option branch. A real run never reached it.

- **Fix.** The line is now the first statement of `main` in all three gated benches.
- **Guard.** `allocator_consistency` checks that. Removing the line from the BI bench fails the new test.

The allocator install itself was correct, as the numbers above show.

PR #1271, opened 2026-09-16
Column promotion moves values instead of copying them (#1269)

sandeepkunkunuru:
Refs #1269. This is the cause of the glibc fragmentation that remains for embedded users after ADR-038.

## Cause, measured

Loading LDBC SF1 on glibc, `mallinfo2` every 100,000 Comment rows (run 35149366139):

| Comment rows | free chunks | free KB | arena MB |
|---:|---:|---:|---:|
| 100 K … 1.8 M | 6–13 | 528–640 | 546 → 1,208 |
| **1.9 M** | **1,775,224** | 129,691 | **1,375** |
| 2.0 M | 1,216,558 | 87,896 | 1,375 |
| end of Comment file | 952,851 | 66,286 | 1,375 |

It is one event, not gradual churn. `ColumnData::maybe_promote` ran at a power-of-two size:
1. it cloned every value of a sparse column into the dense array;
2. it dropped the map, which freed every original.

For a string column that is one small allocation and one small free per entry, all at once, scattered among the copies. It also held two copies of the column at that step (+167 MB arena).

Pre-5b builds had the same ~952 K chunks after the Comment file (run 35146599426). Their per-relationship row maps later refilled the holes, down to 11 after the LIKES relationships. MVCC step 5b removed those maps, so the chunks stayed, and every allocation a query made paid for them: BI-9, BI-12 and BI-14 1.5–1.7× slower on glibc.

## Fix

`maybe_promote` now takes the map out of the column and moves each value into place: no copies, and no per-entry frees.

## Test

`tests/column_promotion_moves_values.rs` uses an allocator local to that test binary. It counts allocations during the insert that promotes a 4,096-entry string column:
- **before the fix:** 4,098 allocations (one copy per entry), so the test fails;
- **after the fix:** fewer than 64 allocations, and the values are intact.

The fixture fills even rows, then odd rows, because a column of contiguous rows promotes at the first check (1,024) instead.

## Results (vm-1, debug)

| run | result |
|---|---|
| `column_promotion_moves_values` | red before, green after |
| `--lib` | 2,239 passed |

A same-runner measurement is running on glibc: main, this fix and pre-5b, compared on free chunks after load and on BI-9/12/14. Results will be posted here.


sandeepkunkunuru:
**Same-runner verification on glibc** (run 35152082412). Three builds:
- main `f8c5fec` with `--no-default-features`, which is what embedded users run;
- this PR, also with `--no-default-features`;
- pre-5b `438493d`.

LDBC SF1, all 20 BI queries, 3 interleaved rounds, median of per-query medians. Ratios are against main.

| | main | this PR | pre-5b |
|---|---:|---:|---:|
| glibc free chunks after load | 797,348 | **199,344** | 219,070 |
| live heap after load | 2,295 MB | 2,293 MB | 3,766 MB |
| BI-14 | 893 ms | **615 (0.69×)** | 657 (0.74×) |
| BI-12 | 1,233 | **806 (0.65×)** | 862 (0.70×) |
| BI-9 | 721 | **504 (0.70×)** | 547 (0.76×) |
| BI-8 | 607 | 517 (0.85×) | 517 (0.85×) |
| BI-4 | 251 | 195 (0.78×) | 205 (0.82×) |
| BI-2 | 5,287 | 4,880 (0.92×) | 4,947 (0.94×) |
| BI-17 | 1,502 | 1,368 (0.91×) | 1,399 (0.93×) |
| BI-1 | 1,029 | 1,079 (1.05×) | 1,070 (1.04×) |

#1269's acceptance criteria:

| criterion | result | met |
|---|---|---|
| free chunks within 1.5× of pre-5b | 199 K against 219 K | yes |
| BI-9/12/14 within 1.1× of pre-5b | faster than pre-5b on all three | yes |
| no memory regression | heap unchanged; step 5b's saving kept | yes |

The worst query is BI-1 at 1.05×. Pre-5b is 1.04× on the same runner, and 1.05× is under CH-REGRESS's 1.1× threshold. The other 11 BI queries are between 0.91× and 1.02×.

PR #1301, opened 2026-09-18
algorithms: write the conventions down, and fix the one that was a hang (ALGO-03)

sandeepkunkunuru:
## What

`docs/ALGORITHM-CONVENTIONS.md`: directedness, weights, self-loops, disconnected
components, tie-breaking and normalisation for all 92 algorithm names the
dispatcher accepts (ALGO-03, previously unmeasured and undocumented).

Plus the defect that writing them found: `CALL algo.maxFlow(n, n)` never
returned.

## The hang

`edmonds_karp` with `source == sink`: BFS pops the source, sees `u == t_idx`,
reports a path. The two walks that follow are `while curr != s_idx`, which with
`curr == s_idx` run zero times — `path_flow` stays `f64::INFINITY` and no
residual capacity changes. The next iteration finds the same path. Forever.

One query hung the thread that ran it. Reproduced before the fix:

```
test flow::tests::max_flow_to_itself_terminates ... FAILED
edmonds_karp did not return within 10s for source == sink
```

`edmonds_karp` now refuses `source == sink`: the flow from a node to itself is
unbounded, so there is no number to report.

## The wrong answer next to it

The operator answered *every* refusal with `max_flow: 0.0`. That is a real flow
value — it means the sink cannot be reached from the source — so a node that is
not in the graph came back as a measurement. Both refusals are now errors.

## How the document is held to the code

`tests/algorithm_conventions_documented.rs` reads the names from
`is_algorithm`'s own match arms rather than from a second list, and fails when
one has no row or has a row with a cell it has not filled in. A cell counts only
when it says something: an em dash, `n/a` or `TBD` reads as undocumented,
because a table of dashes would otherwise score full marks. The second test
checks the gate can fail, which is the way a text check usually stops working.

What this does **not** check is whether a stated convention is *true* —
`CH-ALGO-PARITY` compares answers against a reference, and a doc that says
"self-loops are skipped" over code that counts them is a wrong document, not a
missing one.

## Rows read from the body, not the doc comment

Each convention was read from the implementation. Where a doc comment claimed
something the body did not do, the body won and the mismatch is filed
separately. The document marks with **(differs)** the conventions a user coming
from NetworkX or LDBC would get wrong by assuming — the strongest being that
`mst` silently returns only the component holding internal index 0, and that
`weightedPath` skips negative edges rather than rejecting them.

## Tests

```
cargo test -p samyama-graph-algorithms --lib flow::     2 passed
cargo test --test algo_integration_test                 2 passed
cargo test --test algorithm_conventions_documented      2 passed
```


sandeepkunkunuru:
**This PR now carries #1306 as well.** A merge train I was running had a bug — its wait loop called `systemctl is-active` without `--user`, so it never waited and started merging alongside the first train — and it squash-merged #1306 into this branch rather than into main.

No harm done, and nothing is lost: this branch is now the conventions document and the `maxFlow` hang fix *plus* the negative-weight refusal, which were always meant to land in that order anyway. #1306 is closed as merged. CI has re-run on the combination.

#1307 (the MST forest) still targets `algo/refuse-negative-weights` and will be retargeted to main once this lands. Its content will already be in main by then for the shared files, so the merge is a no-op for those and adds only the MST work.

Flagging it because the PR is now larger than its description, and because the cause is worth knowing: **a merge train that does not wait merges into whatever the base branch happens to be.**

PR #1307, opened 2026-09-18
mst: span every component, and break equal weights the same way twice (#1302)

### LINKED ISSUE #1302: prim_mst silently returns the MST of one component
`prim_mst` takes no start node and returns the spanning tree of whichever component holds internal index 0. On a disconnected graph the result is a spanning tree of one component, `total_weight` is that component's total, and nothing in the result says so. A caller asking for a minimum spanning forest gets a subset with no signal that it is one.

The doc comment describes an API that does not exist:

> returns MST of the component containing `start_node` (or arbitrary node if not specified)

`prim_mst` has no `start_node` parameter. The start is `let start_idx = 0;`, so the component is not the caller's choice and not arbitrary — it is whichever component the dense index 0 landed in, which depends on insertion order.

Found while writing `docs/ALGORITHM-CONVENTIONS.md` (#1301), where it is recorded as a **(differs)** convention.

Two ways out, and the choice is a product one:

1. Return the minimum spanning **forest** — every component — and report the number of components alongside the total. This is what NetworkX's `minimum_spanning_tree` does on a disconnected graph.
2. Keep one component and say which, by returning the component's node count so a caller can compare it against the graph's.

Either is better than the current answer, which is a correct MST of a graph the caller did not ask about.

sandeepkunkunuru:
Stacked on #1306, which is stacked on #1301. Closes #1302.

## What was wrong

`prim_mst` began at internal index 0 and stopped when that component was exhausted. On a disconnected graph the result was a correct minimum spanning tree — of a graph the caller had not asked about. `total_weight` was that one component's total, and nothing in the result said so.

Its doc comment described a `start_node` parameter the function has never had.

## What it returns now

A disconnected graph has no spanning tree, so the answer is the spanning **forest**: one tree per component, `total_weight` summed over all of them, and `components` saying how many trees. `algo.mst` yields that count beside `total_weight`, because the edge rows look identical whether they span one component or five.

On the two-triangle fixture in the new test:

| | before | after |
|---|---|---|
| edges | 2 | 4 |
| total_weight | 3 | 12 |
| components | not reported | 2 |

## The tie-break

`EdgeState` ordered on weight alone, so among equal weights the choice fell to `BinaryHeap`'s internal order — reproducible for one input, arbitrary as a rule. Two graphs differing only in insertion order produced different trees of the same total weight, and nothing said which was expected. Ties now break on target index, then source index.

The test builds the same all-equal-weights K4 six times and asserts the same tree each time. Without the change it is a tree of weight 3 every run and not necessarily the same one, which is exactly the kind of "passes but proves nothing" the assertion is written to avoid.

## Tests

```
cargo test -p samyama-graph-algorithms --lib        97 passed
cargo test --test algo_integration_test              2 passed
cargo test --test algorithm_conventions_documented   2 passed
```

`docs/ALGORITHM-CONVENTIONS.md` loses two **(differs)** marks as a result.


sandeepkunkunuru:
Rebuilt on current main. The branch was `DIRTY` and the cause is worth recording, because I predicted it would be fine and it was not.

#1301 was **squash-merged**, so main gained one new commit with that content. This branch still carried the *original* commits for the same content, so merging main back in put `README.md` and `src/query/executor/operator.rs` in "changed in both" — identical text, different history, a real conflict. A merge does not save you from that; only the content matching by luck would, and it did not.

The straight net diff was not the answer either: this branch predates #1296, #1310 and #1312, so applying `main..branch` wholesale would have **reverted** the SBOM workflows, the REPL, the failure-mode table and the PCA parity work.

So the four commits that are genuinely this PR's were cherry-picked onto main instead. Three applied cleanly; the README one conflicted with the failure-mode link #1310 added to the same table and was re-applied by hand.

What is left is exactly this PR's own work, and nothing else:

    README.md                                    1 +
    crates/samyama-graph-algorithms/src/mst.rs 195 ++++++---
    docs/ALGORITHM-CONVENTIONS.md               13 +-
    src/query/executor/operator.rs               4 +
    tests/algorithm_conventions_documented.rs   57 ++-

Verified locally on the rebuilt branch:

    cargo test -p samyama-graph-algorithms --lib mst::        4 passed
    cargo test --test algo_integration_test                   2 passed
    cargo test --test algorithm_conventions_documented        4 passed

Issue #1309, opened 2026-09-18
ACID_GUARANTEES.md claims fsync-before-ACK and Raft quorum; neither exists


sandeepkunkunuru:
`docs/ACID_GUARANTEES.md` is linked from the README and marks Durability **✅**. Three of its claims describe code we do not have. All verified by grep and by reading the paths, not inferred.

## 1. There is no fsync anywhere on the write path

- `src/persistence/wal.rs:187` — `sync_mode: false, // Default to async for performance`
- `src/persistence/wal.rs:192` — `set_sync_mode` is defined and **never called**. `grep -rn set_sync_mode --include=*.rs .` returns the definition and nothing else, so `sync_mode` is false for the life of the process.
- `src/persistence/wal.rs:232` — even with it true, the call is `file.flush()` on a `BufWriter`. That moves bytes into the OS page cache. It is not a durability barrier.
- `sync_all` / `sync_data` appear in `src/` exactly twice, both in `src/snapshot/persist.rs:50,57` — the snapshot path, not the WAL.
- `src/persistence/storage.rs` sets no `WriteOptions`, so RocksDB's writes are unsynced too.

The document says (line 95) that our write latency is higher than eventual-consistency systems because of "WAL fsync + (in HA) Raft replication before ACK". Neither happens. A committed write is in the page cache when COMMIT replies; a host crash or power loss can lose it.

## 2. Raft is a local apply, and the document claims quorum before ACK

- Line 37: "A write is acknowledged only after a majority of nodes have logged it; this gives linearizability at the cluster level."
- Line 89: "Raft replication ensures data is on a quorum before acknowledgement."
- `src/raft/node.rs:104` — `/// Submit a write request (goes through Raft consensus)`, and the body applies to the **local** state machine and increments `last_log_index`. No log append, no peer contact, no quorum.
- `src/raft/node.rs:47` says so in a comment: "This is a simplified implementation."
- `openraft` supplies a `Config` struct and nothing else.
- No protocol write path calls it: `ClusterManager` in `src/protocol/server.rs` is used for tenant routing and proxying only.

## 3. §1 says write-ahead; the Cypher path is write-behind

Line 30: writes "append to the Samyama logical WAL (ADR-023) **before** in-memory state is mutated."

[...]

Issue #1274, opened 2026-09-17
[REL-03] COMMIT reports success when persistence fails


sandeepkunkunuru:
**Observed in the code at `73e6733`; not reproduced with a failing disk.**

A committed transaction is persisted after the in-memory commit. If that fails, the server logs a warning and still reports success:

| path | code | reply on persistence failure |
|---|---|---|
| RESP `GRAPH.COMMIT` | `src/protocol/command.rs` `commit_transaction_on` | the version, as on success |
| HTTP `POST /api/tx/{id}/commit` | `src/http/transactions.rs` `commit_handler` | `{"committed": true}` |
| RESP single write statement | `src/protocol/command.rs` (`Failed to persist write`) | the query result |

A client told "committed" can lose the write on restart. The in-memory store also keeps the write, so memory and disk disagree until restart.

**What a fix has to decide:**
- Persist before replying, and return an error when persistence fails.
- On failure, either undo the in-memory write (the session transaction's undo log can do this before `commit_session_transaction` clears it) or mark the server read-only until the disk recovers.

**Test first:** a persistence layer that fails `apply_mutations` on demand; assert COMMIT returns an error and the write is not visible after.

Found while writing the REL-01 isolation table (#1273, #1200).

sandeepkunkunuru:
Also affected: **HTTP single write statements.** `AppState::mutate` (`src/http/server.rs`) logs `failed to persist … mutations` and returns the body result. #1275 fixes the transaction COMMIT paths only; single statements have no undo log to roll back with, so they stay open here.

sandeepkunkunuru:
Fixed, with tests on both paths. Verified on `main` at `a9d634c`:

```
$ cargo test --lib a_commit_that_cannot_be_persisted
test http::transactions::tests::a_commit_that_cannot_be_persisted_is_refused_and_rolled_back ... ok
test protocol::server::tests::a_commit_that_cannot_be_persisted_is_refused_and_rolled_back ... ok
```

Both assert what the issue asked for: the COMMIT returns an error rather than the version, and the write is not visible afterwards — in memory **and** on disk, where a write that got through before the failure is put back too.

That covers the first two rows of the table in the issue (RESP `GRAPH.COMMIT` and HTTP `POST /api/tx/{id}/commit`). The third — a RESP single write statement outside a transaction, logging `Failed to persist write` — is worth a separate look if it is still swallowing; I have not checked that path and am not closing it on an assumption.

[...]

PR #1310, opened 2026-09-18
docs: correct the durability and clustering claims (#1309)

sandeepkunkunuru:
Fixes the document half of #1309. Documentation only — no code changes.

`docs/ACID_GUARANTEES.md` is linked from the README and carried a green tick for Durability. Three of its claims describe code we do not have. Every line below was checked against the source before the edit.

## 1. There is no fsync

| | |
|---|---|
| `src/persistence/wal.rs:187` | `sync_mode: false, // Default to async for performance` |
| `src/persistence/wal.rs:192` | `set_sync_mode` is defined and never called — `grep -rn set_sync_mode --include=*.rs .` returns only the definition |
| `src/persistence/wal.rs:232` | even when true, the call is `file.flush()` on a `BufWriter` — the page cache, not a durability barrier |
| `src/` | `sync_all`/`sync_data` appear only in `src/snapshot/persist.rs:50,57` |
| `src/persistence/storage.rs` | no `WriteOptions::set_sync`, so RocksDB is unsynced too |

A committed write survives the process being killed. It may not survive power loss, a kernel panic, or a host reset. The summary row is now ⚠️ and says which is which, rather than ✅ with nothing behind it.

## 2. Raft acknowledges nothing

`src/raft/node.rs:104-119` — the doc comment says "goes through Raft consensus"; the body applies to the **local** state machine and increments `last_log_index`. No log append, no peers, no quorum. Line 47 of the same file calls it "a simplified implementation", and no protocol write path reaches it — `ClusterManager` is used for tenant routing and proxying only.

Both quorum-before-ACK claims are **withdrawn**, not softened. A reader cannot act on a hedge about whether their data is replicated.

## 3. Write-behind, not write-ahead

§1 said writes "append to the Samyama logical WAL **before** in-memory state is mutated". `src/protocol/command.rs:238-256` runs `enable_write_log()` → `execute_mut(...)` → `take_write_log()` → `apply_mutations(...)`, with the append inside `apply_mutations`. A crash in that window leaves nothing to replay, so "Recovery replays from the logical WAL" did not cover the case it was written for.

## Two smaller ones

- **The COMMIT bullet was stale.** It said a failed persist "logs a warning and still reports success". #1275 changed that; the section was written against `73e6733`, one commit earlier. Two tests pin the current behaviour (`a_commit_that_cannot_be_persisted_is_refused_and_rolled_back`, in `src/protocol/server.rs:507` and `src/http/transactions.rs:234`). The *other* half of #1274 — a single write statement outside a transaction — is still true and is now stated separately rather than bundled.
- **§1's "all or nothing"** now says what `src/protocol/command.rs:246-248` already says in a comment: there is no statement rollback (LANG-07), so a statement failing partway keeps the rows it already wrote.

## What §3 got right

The isolation section is in good shape and is untouched: all 15 tests it names exist, and each asserts the anomaly rather than that a call returned `Ok`. The difference is that §3 names tests and §1, §2 and §4 named none — which is how three wrong claims survived in the same file.

## Not fixed here

Whether fsync-before-ACK should be the product behaviour is a decision, not an edit; #1309 carries it, together with the durability level a user would choose and its measured cost.


sandeepkunkunuru:
Added `docs/FAILURE-MODES.md` to this branch — it is the other half of the same problem and the same evidence standard.

REL-15 asks for 20 failure scenarios with observed behaviour, the data guarantee and the operator action. 31 rows are here, each naming a test that exists and asserts something a user can see, and the seven gaps are in the table **as gaps** — disk full, a real SIGKILL, WAL replay after a crash and a corrupt WAL have no test at all (#1311). Every named test was confirmed to exist; five bodies were read rather than trusted by name.

Two rows are deliberately unflattering, and they are the ones this branch exists for: a statement that fails part-way keeps the rows it already wrote, on disk as well as in memory, so the guarantee is that the two agree rather than that the statement was atomic; and power loss after COMMIT has no guarantee at all.

sandeepkunkunuru:
Ran six of the rows through `cargo test --workspace <name>` to check the document's own instruction works, rather than assuming it:

    a_commit_that_cannot_be_persisted_is_refused_and_rolled_back   2 passed (RESP + HTTP)
    a_partial_failure_leaves_disk_agreeing_with_memory             passed
    restore_skips_partial_write_without_marker                     passed
    hostile_parameter_values_cannot_add_a_clause_or_write          passed
    a_connection_that_closes_mid_transaction_rolls_it_back         passed
    an_exploding_operator_is_refused_by_name                       passed

All 36 test names across both documents were also confirmed to exist by grep, and five bodies were read rather than trusted by name.

Issue #1022, opened 2026-08-31
algo.pca: implemented in the algorithms crate, unreachable from Cypher


sandeepkunkunuru:
`pca()` lives in `crates/samyama-graph-algorithms/src/pca.rs` and is re-exported from the crate root and the SDK. From Cypher it does not exist: no arm in the executor's algorithm dispatch, and no entry in `KNOWN_FUNCTIONS`, so `CALL algo.pca(...)` is refused as an unknown algorithm with no hint.

**How it stayed hidden.** CH-ALGO-COV counts algorithms that are *callable*. An implemented-but-unwired algorithm is invisible to a measurement built that way — it can only ever be found by probing a name the implementation does not claim, which is why the probe's candidate list is written from what a user would reach for rather than from the dispatch table. It showed up in the `unknown_to_the_dispatcher` bucket beside `node2vec`, and the two are not the same situation at all.

**Why it is not a one-line wire-up.** `pca(data: &[Vec<f64>], config: PcaConfig)` takes a dense matrix, not a `GraphView`. Every other `algo.*` procedure projects a graph; this one needs a decision about where the matrix comes from — a named list of numeric properties across a label, most likely — and that projection is the actual design work. The other embedding-shaped candidates (`fastRP`, `node2vec`, `graphSage`) would want the same projection, so it is worth settling once.

Not an ALGO-01 blocker: the H1 target of 40 is met without it.

https://claude.ai/code/session_01YMFy59PQM3rQ71fc35Kr8v

PR #1312, opened 2026-09-18
parity: export PCA and the matrix it was computed from (benchmarks#199)

sandeepkunkunuru:
Gives `pca` a parity reference, which is what ALGO-02 has been missing since #1022 made it callable. Paired with benchmarks `55ae258`, which adds the comparison.

## Why ALGO-02 fell without anything regressing

| | 2026-09-05 (`ca48dd1`) | 2026-09-18 |
|---|---|---|
| callable algorithms | 65 | 66 |
| `pca` | unreachable from Cypher (#1022) | callable |
| `algorithms_with_a_reference` | 1.00 | 0.9839 |
| `checks_agreeing` | 1.00 | 1.00 |

Fixing #1022 moved `pca` from a coverage gap to a parity gap. That is worth naming: **a requirement measured as a ratio over what we ship gets harder every time we ship**, and the only defence is that the second requirement is measured at all.

## Why it can have a reference after all

The tempting conclusion was that PCA belongs in `NO_DETERMINISTIC_REFERENCE`, because a component's **sign** is arbitrary and ours canonicalises none — `docs/ALGORITHM-CONVENTIONS.md` records exactly that. But the membership test for that list is "no answer is uniquely correct", and here that is true only of the signs. The **explained-variance ratio** is a property of the data, identical in any basis, and it is the number a user reads.

## The matrix travels with the answer

The features are exported, not recomputed by the comparator. A parity check whose two halves each build "the same" input is partly a check of two input builders, and when it disagrees you cannot tell which one moved — the same reason this exporter already emits the graphs beside the results.

## Measured

numpy's SVD over the centred matrix, against ours, on all four reference graphs:

```
undirected-40   max_abs 1.110e-16
directed-40     max_abs 1.818e-15
undirected-120  max_abs 8.630e-16
dag-40          max_abs 4.191e-15
```

Tolerance 1e-9. With this in, the comparator reports **258 of 258**.

## Found on the way

Two things this work uncovered, neither in this PR:

- `or.solve`'s parity check has not been running. `cost` and `qty` come back absent from the `Item` nodes, so the exporter writes `NaN` and the comparator saw `null`. The cause is not the solver: **`Node::get_property` returns `None` for any property written through Cypher**, because that path stores columnar and the accessor reads the node's inline map. `GraphStore::node_property` returns it. Filed as #1313 with a five-line reproduction.
- That `null` crashed the comparator, which then wrote no report — and the harness read the **previous run's** report, eleven days old, and stamped it with today's commit. Fixed in the benchmarks repo.


sandeepkunkunuru:
Second commit on this branch restores the `or.solve` check, which had silently stopped running for the same reason.

It read the solved quantities with `node.get_property`, which cannot see anything the Cypher write path stored (#1313). Both `cost` and `qty` came back `None`, the export wrote `f64::NAN`, JSON wrote `null`, and the comparator raised on the multiplication — which is what led to the harness reading an eleven-day-old report as today's result.

`store.node_property` sees both stores. One line, and the check runs again:

    or.solve   objective 44.9550 vs LP optimum 45.0000, sum 24.9850 of 25.0   PASS

Comparator now reports **259 of 259, zero failures** — up from the 254 it manages on main once it is allowed to finish, and from the stale 255 it was reporting before.

sandeepkunkunuru:
Third commit: documents the accessor split where a caller will actually see it, and pins the invariant that should hold.

`Node::get_property`'s doc said "Get a property value", which is what every caller believed. It now shows both paths side by side — a property written through Cypher is `None` here and `Some` through `GraphStore::node_property` — and says what the silence cost, which is this PR's own `or.solve` check.

`tests/property_accessors_agree.rs` asserts the invariant rather than the defect: whichever path wrote the property, `node_property` finds it, and a query sees both nodes. If #1313 is fixed so the two accessors agree, the test keeps passing. A test that encoded the asymmetry would have to be deleted in order to fix the bug, which is the wrong incentive.

This PR is now three things, all from the same thread: give `pca` a reference, restore `or.solve`, and make the accessor that broke it hard to use by accident.

PR #1324, opened 2026-09-18
Resolve GDS procedure names, and only where the answer matches (INT-04)

sandeepkunkunuru:
INT-04 asks for `gds.*` names mapped to our algorithms where the semantics match, with the divergences documented. It has stood `unmeasured`.

**Measured first.** `examples/gds_aliases.rs` runs the GDS spellings against a store and counts what resolves. Before: **0 of 16** — measured against the corpus as first written. Two of its entries were wrong (`louvain` and `nodeSimilarity` were called divergent while both are implemented here under their own names) and correcting them took the list to 18; none of the 18 would have resolved before this change either, since every one carries an execution-mode suffix, but that is read from the old canonicalisation rather than measured against it. The `gds.` prefix strip already in the code resolves `gds.pageRank`, which is not a name anyone writes — GDS names carry a maturity namespace and an execution mode.

**After, measured: 18 of 18**, with:

- `alpha.`/`beta.` and a trailing `.stream` dropped; every other mode left on the name so it does not resolve.
- Five explicit renames (`localClusteringCoefficient`→`lcc`, `labelPropagation`→`cdlp`, `spanningTree`→`mst`, `shortestPath.dijkstra`→`weightedPath`).
- `write`/`mutate`/`stats` refused **with the reason**. These write a property or return a summary; we stream. Aliasing them onto the streaming implementation would do a different thing under a trusted name, and "Unknown procedure" pointed the reader at the wrong problem.

`gds.shortestPath.dijkstra` → `weightedPath`, not `shortestPath`: GDS's dijkstra is weighted, ours is an unweighted BFS. The obvious mapping returns plausible numbers for a different question.

**The check that matters more than the count.** A divergent name must not resolve — coverage can't show that, since divergent names are outside the ratio. It caught `triangles → triangleCount` in my own rename table one revision after I added it: GDS's `triangles` lists one row per triangle, ours counts per node.

`docs/GDS-COMPATIBILITY.md` carries the mapping, the refusals and the limit: "the same answer" is asserted from a code read, not measured against a GDS instance. Two entries in the first draft of that list were wrong (`louvain` and `nodeSimilarity` were called divergent while we implement both), so the doc says to treat that column as the weaker half.

7 tests in `tests/gds_procedure_names.rs`. Existing spellings are unchanged.


sandeepkunkunuru:
Follow-up worth recording rather than force-pushing while this is in the merge queue.

`args_for` in the probe passes two arguments for `weightedPath`, which requires three (source, target, weight property). So `gds.shortestPath.dijkstra.stream` is counted as **resolved** on the strength of the error *not* being "unknown algorithm" — which is what the probe measures and says it measures, but it means that one call never actually executes.

That is sound for INT-04, which asks whether the name reaches the implementation. It is weaker than the neighbouring `the_streaming_mode_actually_runs` test, which executes `gds.pageRank.stream` for real. Passing a weight property would let the dijkstra case be executed too, and would close the gap between "the name resolves" and "the call works" for the one mapping in this PR most likely to be wrong.

Verified separately while reviewing: `execute_shortest_path` takes two arguments and `execute_weighted_path` three, so the claim that GDS's weighted dijkstra maps to `weightedPath` rather than `shortestPath` holds.

PR #1352, opened 2026-09-19
Kill -9 the server mid-write and check what survives (REL-04)

sandeepkunkunuru:
REL-04 asks for `kill -9` at 1,000 random points during a mixed workload, with recovery landing on some prefix of the commit log **every time**. The engine's own status column said "ad-hoc" and the harness reported it `unmeasured`, so nobody knew either way.

### What counts as a failure

| | counted? |
|---|---|
| unacknowledged write lost | no — the client never heard it land |
| **acknowledged write lost** | **yes** — the server answered and the data is gone |
| write in flight when the signal arrived | no — neither acknowledged nor refused |
| write present that was never acknowledged | yes, separately |
| surviving set with holes in it | yes, separately |

That last one matters: "43 of 50 survived" is consistent with a prefix and also with 43 scattered survivors — a state no ordering of the commit log can produce, and a worse failure than losing the tail. The surviving set is compared against the longest prefix of the issue order.

Excluding in-flight writes is not leniency. Counting them would report a failure on every cycle by construction, which is the fastest way to make a real one invisible.

### The kill lands mid-write

The first version of this loop wrote N times and then killed, so every crash happened in the gap between two requests — the one moment the server has nothing half-done. The writer now runs on its own thread and the signal fires after a random delay.

### An acknowledgement is not one thing

`auto` takes the 200 on the statement. `tx` takes the 200 on `POST /api/tx/:id/commit` — the acknowledgement #1275 is about. Different promises, so half the cycles run each and the results are reported separately: a combined figure lets a clean half hide a broken one.

### Break-tested

Against an `--ephemeral` server, which persists nothing:

```
acked 13, survived 0, acked_lost_count 13, longest_acked_prefix_intact 0
```

Without that the probe could not tell loss from no loss.

It is a script rather than a test because `kill -9` has to hit a real process at a real moment; an in-process test can only simulate the crash it is trying to observe. `CH-CRASH` in the benchmarks repo reads its JSON.

A 1000-kill-point sweep is running now; its result goes on the issue rather than into this description, so the number and the code that produced it are not claimed from the same commit.


sandeepkunkunuru:
The sweep is done: **1000 kill points, 905,907 acknowledged writes, none lost, no gapped recoveries** — measured in #1355, with the two things the first two attempts got wrong written up there.

The second commit on this branch is one of them: the prefix was measured over acknowledged ids, so a write that was in flight at the kill and reached disk anyway — correct, and it happened 176 times — pushed the survivor count past the acknowledged prefix and the cycle was reported as gapped. 156 of 997. A check that fires on correct behaviour one time in six gets switched off, and the failure it exists to find goes with it.

PR #1358, opened 2026-09-19
Errors point at the place in the query that caused them (LANG-12)

sandeepkunkunuru:
LANG-12 asks for "codes **and** spans on 100%". Codes were at 100% and spans at **3 of 16** — and all three were grammar errors, where `pest` draws the caret for us. Every semantic and runtime error named the offending variable or function in prose and gave no position.

```
[Samyama.ClientError.Statement.VariableNotBound] `m` is not bound: nothing in this query defines it
 --> 1:18
  |
1 | MATCH (n) RETURN m.name
  |                  ^
```

### Where the span is attached

`QueryEngine` — the layer HTTP, RESP, the CLI and the embedded SDK all go through, and the one that has the query text. **Both of its paths needed it**: semantic checks like an unbound variable or an unknown function are raised by the *parser*, so a hook on execution alone missed exactly half the errors a user meets.

Threading a span through the AST and every error site is the better long-term answer. This is not that; it makes the error useful now, from the token the message already names.

### It annotates only when it finds the token

A span on every error is easy to fake and helps nobody. So four messages were changed to name what they are about — the fix a reader wants anyway:

| before | after |
|---|---|
| `Add requires numeric or string operands` | ``​`+` requires numeric or string operands`` |
| `Binary op requires property values` | ``​`+` needs a value on the left, and got a node, relationship or path`` |
| `an aggregate function is not allowed in WHERE` | ``…in `WHERE`…`` |
| `Cannot execute write query with read-only executor. Use MutQueryExecutor instead.` | ``​`DELETE` writes, and this connection is read-only. Send the statement on a write path instead.`` |

That last one named two internal Rust types at somebody holding a Cypher string and an HTTP endpoint.

`Query::write_clause` had to read the legacy `delete_clause`/`set_clauses` fields as well as `clauses`, or `MATCH (n) DELETE n` reported no write clause at all — the two-AST-shapes trap again.

### The extractor takes the earliest candidate, and only a qualified name after a colon

`Unknown algorithm: algo.nope. Available: pageRank(), wcc()` matches the call shape at `pageRank` and the colon shape at `algo.nope`. **The caret has to land on the fault, not on the advice.**

And a bare word after a colon is prose: `Type error: Add requires numeric` offered `Add`, a Rust variant name — harmless only because `Add` is not in the query. A message reading `Type error: Integer expected` against a graph with a property called `Integer` would have pointed at it with full confidence.

### Two probes were measuring an internal pair

`error_quality` and `error_uniformity` both called `parse_query` + `QueryExecutor::execute` directly. That made no difference until spans existed — and then `error_uniformity`'s embedded arm was the only one without one, and **API-03's message agreement fell to 2 of 13 over a difference that existed nowhere but in the probe**. `EmbeddedClient` holds a `QueryEngine`; the probe now does too.

### Measurement

`CH-ERR` fully green: spans **0.1875 → 1.0**, codes 1.0, both API-03 agreements 13 of 13. 12 unit tests on the extractor and renderer, including the two ways it would lie.


sandeepkunkunuru:
The API-16 corpus gate caught this on the first change that touched the contract, and caught **only** it — 5 of 98 interactions, all `error_*`:

```
CHANGED error_unknown_function
CHANGED error_unknown_procedure
CHANGED error_unknown_algorithm
CHANGED error_undefined_variable
CHANGED error_gds_write_mode
98 interactions replayed, 5 changed
```

Each now carries the caret appended to the same message with the same status code, so the corpus is re-recorded deliberately — which is what `--i-am-changing-the-contract` exists to make someone type.

`error_type_mismatch` (`RETURN 'a' + [1]`) did not move: `+` on a string and a list prepends, so it never reaches the arm whose message changed.

Issue #1365, opened 2026-09-19
RETURN toUpper() aborts the server process — 69 function/arity pairs panic on index out of bounds


sandeepkunkunuru:
```
$ curl -X POST localhost:8080/api/query -d '{"query":"RETURN toUpper()"}'
(no response; the process is gone)

thread 'tokio-runtime-worker' panicked at src/query/executor/operator.rs:2853:37:
index out of bounds: the len is 0 but the index is 0
```

Fourteen characters, over an HTTP endpoint that reads no credential (#1328), and the server is down. With no `--data-path` configured that takes the entire in-memory graph with it.

### It is a class, not a case

A sweep of every name in `KNOWN_FUNCTIONS` against 0–3 arguments found **69 (name, argument-count) pairs that panic** — 67 at zero arguments, plus `atan2` and `hasLabels` at one:

```
abs/0, acos/0, asin/0, atan/0, atan2/0, atan2/1, ceil/0, cos/0, ..., toUpper/0,
trim/0, type/0, valueType/0
```

Roughly 97 places in `operator.rs` index `args[0]` directly. A handful of functions check their arity (`hierarchy_rollup`, `substring`); most do not.

### Fix

A `MIN_ARITY` table checked once before dispatch, rather than a guard at each of the 97 sites — the next function somebody adds will index its arguments too. `tests/function_arity.rs` calls every known name with zero to three arguments, in both the lower-case spelling and a capitalised one, so a new function that needs a row and lacks one fails in CI.

**The case pair in that test is not decoration.** The first version of the guard looked the name up case-sensitively and the first version of the sweep used `KNOWN_FUNCTIONS`'s lower-case spelling. It passed, while `toUpper()` — the spelling everybody writes — still aborted the process.

sandeepkunkunuru:
Fixed in `26ffa64` (#1366). `tests/function_arity.rs` sweeps every name in `KNOWN_FUNCTIONS` against 0–3 arguments and eight argument types, in both spellings — the type variation is what caught `left('a')` and `right('a')`, which a sweep passing `1` everywhere reported as safe. A function added later that indexes its arguments without a `MIN_ARITY` row fails there rather than in production.

Issue #1328, opened 2026-09-18
The HTTP API has no authentication and accepts any origin, and we echo the Private Network Access opt-in


sandeepkunkunuru:
Measured by `CH-SEC` (samyama-graph-competitor-benchmarks). REL-08 is 0 of 4 controls, REL-09 is 0 of 2.

## The reachable state

Three facts about `src/http/server.rs` combine:

1. `POST /api/query` executes arbitrary Cypher, including `DELETE`.
2. Nothing reads a credential off the request. There is no credential store, no `Authorization` handling, no `AUTH`/`ACL` command on RESP, no grants, no audit log.
3. `CorsLayer::permissive()` accepts any origin, and `allow_private_network` echoes `Access-Control-Allow-Private-Network: true` to whichever origin asks for it.

Point 3 is the part worth reading twice. The Private Network Access header exists so that a **public web page cannot reach a private address** unless the local service opts in. The docstring explains the need correctly — the hosted Studio calling a loopback container (#342) — but the implementation opts in for *every* requesting origin, not for the Studio. That converts a browser protection into a browser permission.

So a user running the documented `docker run` and then browsing the web is relying on nothing.

## Not in scope of this issue

I am not including a proof-of-concept. The class is CSRF / DNS-rebinding against a local database, and the shape above is enough to act on.

## Tenant isolation is a naming convention

The tenant is a parameter the caller supplies. `TenantManager` quotas and per-tenant column families genuinely separate data — they do not stop a caller naming a tenant that is not theirs, because nothing establishes who the caller is. Worth stating explicitly, because "multi-tenancy" reads as an isolation control in the feature list.

## Suggested order

1. **Restrict the PNA echo and CORS to a configured origin allowlist**, defaulting to none. This is the smallest change and it closes the browser path on its own.
2. **Bind to loopback by default**, with an explicit flag to listen on a routable address.
3. Then the REL-08 work proper: a credential store, a check on the request path, graph-level grants, an audit log of writes and admin actions.
4. REL-09: a TLS acceptor on our own listeners (the `rustls` already in the tree is used by outbound clients), and at-rest encryption for snapshots.

## What the measurement can and cannot say

`CH-SEC` reads the server's request path with comments stripped. It can say a mechanism is absent. It cannot say a present one is correct — a credential check that exists may still be bypassable, and that needs a test that attacks it rather than a test that reads it.

[...]

PR #1366, opened 2026-09-19
`RETURN toUpper()` aborted the server; 69 calls did (#1365)

sandeepkunkunuru:
```
$ curl -X POST localhost:8080/api/query -d '{"query":"RETURN toUpper()"}'
(no response; the process is gone)

thread 'tokio-runtime-worker' panicked at operator.rs:2853:37:
index out of bounds: the len is 0 but the index is 0
```

Fourteen characters, over an HTTP endpoint that reads no credential (#1328), and the server is down. With no `--data-path` configured that takes the entire in-memory graph with it.

### It is a class, not a case

A sweep of every name in `KNOWN_FUNCTIONS` against 0–3 arguments found **69 (name, argument-count) pairs that panic** — 67 at zero arguments, plus `atan2` and `hasLabels` at one. **Fixing `toUpper` would have left sixty-eight.**

Roughly 97 places index `args[0]` directly. So: a `MIN_ARITY` table checked once before dispatch, rather than a guard at each site — the next function somebody writes will index its arguments too, and a table plus a sweep is the only version of this that stays fixed.

`tests/function_arity.rs` calls every known name with 0–3 arguments and asserts that **nothing aborts**. An error is a fine outcome; most of those calls are nonsense and should be refused. The only unacceptable answer is a panic.

### The two spellings in the sweep are not decoration

The first version of the guard looked the name up case-sensitively, and the first version of the sweep used `KNOWN_FUNCTIONS`'s lower-case spelling. **It passed, while `toUpper()` — the spelling everybody writes — still aborted the process.** The lookup now lowercases and the sweep tries both.

### Verified

Break-tested by disabling the guard: all 138 spellings come back as offenders. Against a live server:

```
RETURN toUpper()          -> ArgumentError: `toUpper()` takes at least 1 argument; got 0
RETURN atan2(1)           -> ArgumentError: `atan2()` takes at least 2 arguments; got 1
RETURN toUpper('ab') AS x -> {"records":[["AB"]]}
/api/status               -> healthy
```


sandeepkunkunuru:
Two more, found by widening the fuzz: **`left('a')` and `right('a')` also abort the process.** They take two arguments and the table had them at one.

The first sweep passed `1` for every argument and reported them safe — `left(1)` is refused on the *type* before anything reaches `args[1]`, while `left('a')` gets that far. The committed sweep now varies the value as well as the count: integer, string, boolean, null, list, map, float, negative, across 0–3 arguments and both spellings.

Issue #1308, opened 2026-09-18
algo.bridges reports parallel edges as a bridge; the fix needs a convention first


sandeepkunkunuru:
Measured, not read. Two graphs through `algo.bridges`:

```
a = b  (two parallel LINK edges), plus b -> c
   bridges = [(1, 2), (2, 3)]     <- (1,2) is wrong

x <-> y  (a reciprocal pair: x->y and y->x)
   bridges = [(1, 2)]             <- correct
```

Removing either of the two parallel edges leaves the graph connected, so `a = b` is not a bridge. It is reported as one. `b -> c` is a bridge and is correctly reported, so the failure is specific to parallel edges rather than to the search.

The comment in `traversal.rs` describes the fix that is missing:

> Only *one* edge to the parent may be ignored: with parallel edges `v == parent` twice, the second is a genuine back edge and the pair is not a bridge. Comparing on node identity alone would call it one.

The body is `} else if w != parent {` — node identity, which is what the comment says would be wrong. `articulation_points` shares the same traversal and the same defect.

## Why it is not a two-line fix

`undirected_neighbours` is `successors ++ predecessors`, so **both** of these put the neighbour in the list twice:

- two parallel `a -> b` edges (not a bridge — skipping one occurrence is right)
- a reciprocal `a -> b`, `b -> a` pair (one undirected edge, and a bridge — skipping both is right)

The list has no edge identity, so nothing in the traversal can tell the two apart. Skipping the parent exactly once — the textbook fix — would fix the parallel case and *break* the reciprocal one, which is the more common shape in our data.

So this needs a convention decided before it needs code: **under an undirected reading, is a reciprocal pair one edge or two?**

- *One edge* matches how a user draws the graph, keeps `x <-> y` a bridge, and means the parallel case needs real edge identity to fix.
- *Two edges* matches Neo4j GDS's undirected projection, makes `x <-> y` not a bridge, and is a behaviour change to every algorithm that symmetrises — `bridges`, `articulationPoints`, `countTriangles`, and everything going through `neighbours()`.

Whichever is chosen belongs in `docs/ALGORITHM-CONVENTIONS.md`, which has no column for it today because the question had not been asked. The fix then follows from the decision.

Found while writing that document (samyama-graph#1301); the comment mismatch is one of the eighteen in #1305.

sandeepkunkunuru:
A concrete convention to decide between, since the issue correctly stops at "this needs a convention first".

[...]

PR #1368, opened 2026-09-20
Parallel edges are not a bridge, and a reciprocal pair is one edge (#1308)

sandeepkunkunuru:
`algo.bridges` reported `a = b` — two parallel `LINK` edges — as a bridge. Removing either leaves the graph connected, so it is not one. The search skipped **every** edge back to the DFS parent on node identity, rather than the single edge it arrived on.

### Why this sat open

The textbook fix breaks the commoner case. `successors ++ predecessors` puts the parent in the list twice for two different reasons, and carries no edge identity to tell them apart:

- two parallel `a -> b` edges — **two** undirected edges, so skipping one occurrence is right;
- a reciprocal `a -> b`, `b -> a` pair — **one** undirected edge that genuinely *is* a bridge, so skipping both is right.

So the convention comes first. It is now in `docs/ALGORITHM-CONVENTIONS.md`: **a reciprocal pair is one undirected edge**, so multiplicity is `max(forward, reverse)`.

| forward | reverse | undirected |
|---|---|---|
| 1 | 0 | 1 |
| 1 | 1 | 1 — a reciprocal pair |
| 2 | 0 | 2 — parallel |
| 2 | 1 | 2 — one pair collapses, the leftover stands alone |

**That is the conservative reading**: it is what the traversal already did for reciprocal pairs, so `x <-> y` stays a bridge and no other algorithm's answer moves. The alternative — a reciprocal pair counting as two, which is Neo4j GDS's undirected projection — would change `bridges`, `articulationPoints`, `countTriangles` and everything reached through `neighbours()`. Recorded in the conventions document as the road not taken rather than left implicit.

With the multiplicity right, skipping the parent exactly once is correct, and that is the code change.

### Tests pin both halves

Nine of them, because **a change that fixed the parallel case by breaking `x <-> y` would pass a test for either one alone.** Break-tested twice:

| break | fails |
|---|---|
| skip the parent on node identity | 3 tests — all the parallel cases |
| count `forward + reverse` instead of `max` | `a_reciprocal_pair_is_still_a_bridge` |

`algo_parity_check`: **62,481 of 62,481** checks still agree with the recorded answers. `articulationPoints` shares the traversal and is pinned with it.


sandeepkunkunuru:
Two follow-up commits, both from the conventions gate catching this before CI did — and I raised the PR before reading the workspace run, which is the wrong order.

**The gate read my new table as broken algorithm rows.** It latched onto the `Cypher name` header and then treated every later pipe row as an algorithm row, so any second table in the document looked malformed. A markdown header is only a header because a `|---|---|` separator follows it, so the parser now looks one line ahead: a header that is not the algorithm header ends the algorithm table. The shift check that caught the real defect — three rows whose columns had slid behind an unescaped `|` — is untouched, and its own break-test still passes.

**And the new section had to move below the connectivity table.** Putting it inside orphaned the biconnected-components row, and the gate correctly reported it as an algorithm with no conventions.

Workspace suite green, `doc_check` 0 failures, `algo_parity_check` 62,481 of 62,481.

PR #1369, opened 2026-09-20
`date('2024-05-06').year` answers, like `d.year` always did (LANG-16)

sandeepkunkunuru:
Component access on a **bound variable** worked. The same access on an **expression** went through the executor's index operator — which knows about lists and maps — and was refused:

```
WITH date('2024-05-06') AS d RETURN d.year   ->  2024
RETURN date('2024-05-06').year               ->  cannot index Date: it is not a list or a map
```

Two spellings of one question, on two code paths, with only one of them implemented.

### One entry point

`temporal_property` is now the single function and both paths call it. The legacy millisecond `DateTime` is read as the UTC zoned datetime it is — lossless, since milliseconds fit exactly — rather than given a second implementation of `.year` that could disagree with the first.

### The tests assert the two spellings *agree*

Not that each works. **A second implementation returning a plausible-but-different answer would pass a test for either one alone.** Every component of `date`, `datetime` and `localtime` is compared both ways, including one the type does not have, where both must answer null.

Three more pin what must not move: an integer and a string still have no components and still say so; a list still indexes by position; a map still indexes by key — the new arm sits on the path they share.

Break-tested by making the new arm return `None`: six of the nine fail.

**LANG-16: 8 of 9 → 9 of 9.** API-16 corpus replays 256 of 256 unchanged; `doc_check` clean.


sandeepkunkunuru:
CI caught this and was right: `CH-API-STABILITY` failed on `date_component`.

```
CHANGED date_component
  .response.error: "cannot index Date: it is not a list or a map" -> absent
  .response.records: absent -> [[2024]]
  .status: 400 -> 200
```

`RETURN date('2024-01-15').year` was recorded when it was a 400, so **the corpus was pinning the behaviour this PR exists to change**. One interaction moved, it is the one that was supposed to, and nothing else in 256 did. Re-recorded deliberately.

My local run was green only because it predated the merge of main that brought the corpus up to 256 — the gate saw what I had not.

PR #1378, opened 2026-09-20
Point it at your queries: a compatibility report (INT-11)

sandeepkunkunuru:
## The check that could not pass

INT-11's measurement was the literal `False`, with a note explaining that `docs/CYPHER_COMPATIBILITY.md` answers a different question. That was true, and it could not become false however much tooling shipped.

The matrix says which features exist. Somebody deciding whether to migrate is asking whether **their** queries run, and a list of 78 supported features says nothing about the twelve queries in their application that use the two that are missing.

## The tool

```bash
cargo run --release --example compatibility_report -- --queries your_queries.cypher
```

Reports what is accepted, what is refused, and why — **grouped by cause**. A migration is planned against causes: six hundred failures from one missing function is an afternoon; six failures from six causes is six.

Against `benchmarks/compat/neo4j-idioms.cypher`, 59 queries:

```
  accepted       53  (89.8%)
  refused        6
       1x  NotFound      Unknown procedure: apoc.periodic.iterate
       1x  SemanticError CALL … YIELD … WITH  (#1375)
       1x  SyntaxError   expected EOI, …      (USING INDEX — LANG-13)
       1x  SyntaxError   expected EOI, …      (apoc.text.join in RETURN)
       1x  SyntaxError   expected label       (a backticked label — #1373)
       1x  SyntaxError   expected labels, properties, or variable  (a backticked variable — #1373)
```

## The corpus is written from the outside

From Neo4j's documentation and from application code, **not** from our feature matrix — one derived from the matrix would report 100% of whatever we already do.

Writing it that way immediately caught one of my own entries being invalid Cypher: `MATCH (p:Person) RETURN p UNION MATCH (c:Company) RETURN c`, which Neo4j rejects too because UNION requires matching column names. It was inflating the refusal count with our mistake, so the file now says entries must be valid Cypher and says why.

## "Accepted" is a narrow claim

It parses and plans. That is about the language, not about the data and not about the answer: an accepted query can still return the wrong rows and this tool would not know. It is the strongest thing sayable without the user's database; a percentage claiming more would be a guess wearing a number. The report says so in its own output.

## Where the logic lives

`src/compat.rs`, not the example — so the report and the test that pins it cannot drift. Two copies of a classifier is how a report starts disagreeing with the suite that checks it.

`tests/compatibility_corpus.rs` is a **ratchet, deliberately asymmetric**: a query that used to be accepted and is now refused fails; one that used to be refused and is now accepted passes, because improvement should never break a build. A ratchet that reads zero is worse than none, so the first assertion is that the corpus was found and split into fifty-odd queries at all. Break-tested — adding one unsupported idiom fails it by name.

## Also

`docs/MIGRATING-FROM-NEO4J.md`: what each cause usually means, the two data routes that work, and the four things worth knowing before committing — including that there is no authentication.

Workspace suite green.


sandeepkunkunuru:
Numbers moved after #1376 (backticked labels) merged and this branch was updated: the corpus now reads **54 of 59 accepted, 5 refused** — `MATCH (:`Research Paper`)` is accepted. The ratchet's floor is raised to 54 in this branch and that query's line is removed from `KNOWN_REFUSALS`.

Worth noting what the merge exercised: the ratchet passed *before* the floor was raised, because refusals are matched by query prefix rather than by error code — and `apoc.text.join` had changed from a syntax error to an unknown-function error (#1386) without changing whether it is refused. An error-code match would have failed there for no reason.

The body above and `docs/MIGRATING-FROM-NEO4J.md` both carry the new figure.

PR #1386, opened 2026-09-20
A namespaced function name may have more than one segment

sandeepkunkunuru:
## The problem

`function_name` allowed **one** namespace segment — `date.truncate`, `duration.between` — so `apoc.text.join(...)`, two segments and the commonest shape in a Neo4j corpus, was a parse error:

```
expected EOI, union_clause, order_by_clause, skip_clause, limit_clause…
```

A user migrating from Neo4j learns nothing from that.

## After

The name is accepted to any depth, the query reaches the evaluator, and the answer is:

```
Unknown function: apoc.text.join
```

which names the thing they have to replace.

**Accepting a name is not implementing it**, and the refusal is the point — it is a *better refusal*. Found by pointing `examples/compatibility_report` (#1378) at a corpus of Neo4j idioms, where this was one of six causes and the only one whose message identified nothing.

## The risk, and the evidence

`function_name` is greedy over dots and `d.doc.a.b.c` is property access. They do not collide because `function_call` requires a `(` immediately after the name, so a dotted path with no call backtracks to `property_access`.

That is an argument, so the tests lead with the evidence rather than the feature: a **four-segment** nested map read in `RETURN` and in `WHERE` — `WHERE` parses expressions by a different route, so it gets its own case rather than being assumed to follow. A regression there would break every nested property access in the engine.

`duration.between` is checked too: it was implemented and unreachable from Cypher until one segment was allowed (#769), and widening the rule must not take it away again.

Workspace suite green.


sandeepkunkunuru:
After #1376 merged, `apoc.text.join` is one of **five** remaining causes in the corpus rather than six — the backticked-label cause is closed. It is still the only one whose message identifies nothing, which is what this change fixes.

PR #1397, opened 2026-09-21
NDS-06 + NDS-07: full-text search — tokenised, stemmed, BM25-scored

sandeepkunkunuru:
`CONTAINS` was the only text search here, and it is a substring match. Four separate problems, of which only the last is about speed:

| | `CONTAINS 'graph'` | this index |
|---|---|---|
| `Graphs` | misses it | finds it |
| `polygraph` | **matches it** | does not |
| ranking | none | BM25 |
| cost | scans every node | term lookup |

```cypher
CREATE FULLTEXT INDEX docs FOR (d:Doc) ON (d.body);
CALL db.index.fulltext.queryNodes('docs', 'graph databases')
YIELD node, score RETURN node.title, score ORDER BY score DESC;
```

## Measurement

**NDS-06 1/3 → 3/3 and NDS-07 2/3 → 3/3 — both met**, by `examples/nds_probes`. NDS-07's missing probe was "text search composed with a graph traversal", which is the `CALL … YIELD node MATCH (node)-[]->(m)` form and has a test of its own.

## Five decisions, each a way to be quietly wrong

**IDF is floored at zero.** Unfloored, a term in more than half the corpus scores negative and a document is pushed *down* the ranking for containing a word the user searched for.

**A phrase checks positions, not co-occurrence.** Checking that both words are present matches a document holding them a page apart — and looks right in every small test.

**The stemmer is not Porter, and the docs lead with that.** Fifteen listed suffix rules with a minimum stem length, which is what stops `is` becoming `i` and collapsing unrelated words into one token. The property that actually has to hold is that a query is stemmed by the same function that stemmed the document; applied to one side only it silently stops matching the very words it was added for.

**No stopword list, deliberately.** BM25 already gives an everywhere-word zero weight, and removing such words breaks every phrase containing one — `"shortest path to the node"` has two.

**Maintenance hangs off `apply_property_set`**, the same choke point as the property and vector indexes. Maintained from that function's callers instead, whichever one got forgotten would go stale — and a stale inverted index reports nothing at all: it returns a document that no longer holds the word, and the caller cannot tell that from a correct hit. `CREATE` backfills (the vector index's `create_vector_index_backfills_existing_embeddings` bug, avoided); a `SET` replaces the document's terms; a property that stops being a string is dropped.

## Tests

23 (12 unit + 11 through Cypher). Break-tested:

| Break | Tests that fail |
|---|---|
| disable the write-path hook | 2 |
| disable the backfill at `CREATE` | 7 |

**An existing guard caught a real defect in this change.** `every_mutator_bumps_the_epoch_or_is_listed` failed: creating an index changes what `queryNodes` returns — from an error to rows — so a result cache keyed on the epoch would have served the old answer. Fixed by bumping the epoch, not by adding the method to the exemption list.

## What it does not do

Named in `docs/FULL-TEXT-SEARCH.md` rather than left to be found: relationship properties are not indexed; `ON EACH [a, b]` creates one index per property rather than Neo4j's single index over two fields (so cross-field scores are not comparable — stated, not glossed); nothing is persisted, so the DDL must be re-run after loading a snapshot; there is no language configuration and CJK text does not tokenise usefully; and `SHOW INDEXES` does not list full-text indexes yet, so the error from an unknown index name — which lists the known ones — is currently the only discovery path.


sandeepkunkunuru:
**Integration check across the seven open branches.**

Per-PR CI tests each branch against `main`, never against the others. With #1390, #1391, #1392, #1394, #1395, #1396 and this one all merged into one tree off `main` (`4b49456`):

- merges clean — `src/lib.rs`, `src/query/cypher.pest` and `src/query/executor/operator.rs` all auto-merge
- `cargo test --lib`: **2287 passed, 0 failed**
- `cargo test --tests`: **347 test binaries, all ok**

Two textual conflicts were predicted and pre-empted rather than left for the merge train: `pub mod provenance;` moved off the line #1392 also edits, and this PR's grammar rules moved away from the `create_vector_index_stmt` lines #1391 rewrites. Adjacent hunks conflict in git even when the changes are unrelated.

Requirement probes on the combined tree (`examples/nds_probes`): NDS-05, **NDS-06**, **NDS-07**, NDS-08 and NDS-13 all at full marks. NDS-11 stays 5/8 — `CREATE FULLTEXT INDEX ON :N(x)`, the unnamed plain spelling, is still not accepted, and is not addressed here.

Issue #1327, opened 2026-09-18
The arm64 image is built and never started, and releases carry no binaries (DX-03)


sandeepkunkunuru:
Measured by `CH-DEPLOY` (samyama-graph-competitor-benchmarks). INT-12's H1 bar is met; DX-03's is not, in two ways.

## The arm64 image has never been run

`publish-docker.yml` builds `platforms: linux/amd64,linux/arm64`, and the smoke-test job runs on `ubuntu-latest` — amd64. So the arm64 image is pushed to GHCR without anything ever having started it.

The smoke test itself is good: it starts the container, waits for `State.Health.Status`, curls `/api/status`, and runs a `GRAPH.QUERY` over RESP. It just only ever does that on one architecture.

DX-03 asks for the one-liner "verified in CI on 2 architectures". Built for an architecture is not verified on it — an image that fails to start on arm64 would ship and nothing would notice until a user pulled it.

Fix: run the same smoke job on an arm runner (`ubuntu-24.04-arm` is already used by the test matrix in `ci.yml`), pinned to the same version tag.

## No single-binary download

DX-03's other half is "a single-binary download … with zero configuration". `v1.8.0` carries **zero release assets** — no binary for any platform.

```
gh release view v1.8.0 --json assets  ->  {"assets": []}
```

`tag-release.yml` uploads an SBOM, but no compiled binary, and in any case nothing was attached to v1.8.0 (see #1249/#1250 for why the release workflow produced nothing).

Fix: build and attach `linux/x86_64` and `linux/aarch64` binaries at tag time, and add a step that downloads the published asset and runs `--version` — otherwise this repeats the pattern the check was written to catch, where a publish step stands in for a published artifact.

## Not in scope here

No Helm chart exists (`Chart.yaml` is absent anywhere in the repo). INT-12 asks for one at H2, so it is reported and not counted against H1.

Issue #1038, opened 2026-09-01
The published Docker image (1.1.0) cannot parse MATCH (a)-->(b); the engine is at 1.7.1


sandeepkunkunuru:
Measuring DX-01/DX-02 against the image the README's quickstart tells everyone to pull:

```
docker pull public.ecr.aws/f9f6l5u4/samyama-graph:1.1.0
POST /api/query  {"query":"MATCH (a)-->(b) RETURN count(*) AS n"}
→ {"error":"Parse error: --> 1:11 ... = expected edge_detail"}

POST /api/query  {"query":"MATCH (a)-[]->(b) RETURN count(*) AS n"}
→ {"records":[[38539]]}
```

The bare arrow is ordinary Cypher and about the simplest pattern there is. It was fixed in `src/query/cypher.pest` on 2026-08-26 (`25bef2d`) by making `edge_detail` optional — the comment there says it plainly:

> Requiring `edge_detail` made `MATCH (a)-->(b)` — about the simplest pattern there is — a syntax error, which is invisible in a codebase whose own queries always name a type.

That last clause is why this survived to a release and why the README does not trip over it: **every Cypher example in our own docs names a relationship type.** A new user writing textbook Cypher does not.

**Cargo.toml is at 1.7.1. The published image is 1.1.0.** Anyone following the quickstart today gets an engine six minor versions behind, and the first thing they are likely to type fails with a syntax error against valid Cypher. Whatever the TCK says about the code, this is what the artifact does.

Everything else on the documented path works and is fast — full cold pull → snapshot import → answered multi-hop query is **18 seconds** against DX-02's 15-minute target, and 9–12 s to a first query (DX-01). The path is good; the image on it is old.

Asks:
1. Publish an image built from current `main`, and say in the README which engine version the tag corresponds to.
2. Add the image tag to whatever gates a release, so `1.7.1` cannot ship while `1.1.0` is what `docker pull` returns.

Related: #1037 (the quickstart never shows how to run a query at all).

PR #1405, opened 2026-09-21
ci: mirror the published image to ECR Public, instead of pushing it by hand

Kalyanpulikanti-01:
ECR Public holds `1.1.0`, `1.7.0` and `1.7.1`. All three were pushed by hand — a search of this repository for `ecr` or `f9f6l5u4` across workflows, scripts and docs returns **nothing** — and **v1.8.0 is absent** because nobody ran the manual step. Nothing reported that it had been skipped.

The tags that *were* pushed went unused. Every reference across the KG repos and `samyama-cloud` still names `1.1.0` — 70 of them across 11 repos (`assetops-kg`, `clinicaltrials-kg`, `cricket-kg`, `druginteractions-kg`, `edtech-kg`, `legal-judgments-graph-kg`, `mental_health-kg`, `pathways-kg`, `regulatory-affairs-kg`, `samyama-cloud`, `samyama-graph-web`). Not one mentions 1.7.0 or 1.7.1. So the manual step was both unreliable and invisible when missed, and the image new users are told to pull predates the `MATCH (a)-->(b)` fix (#1038).

## Shape

A separate `mirror-to-ecr` job that runs **after** `build-and-push` and `smoke-test`, and copies by digest with `imagetools create`. Three properties it is built for:

1. **It cannot break the GHCR publish.** A failed assume-role or an AWS outage costs the mirror, not the release. Adding ECR to the existing `build-push-action` step would have made a working publish depend on a registry that has never been automated.
2. **It copies, it does not rebuild.** ECR serves the exact bytes and the same digest GHCR does. A second build would produce a second image that merely *ought* to be identical.
3. **It waits for the smoke test.** #1327 was an arm64 image that had never been started reaching a registry. Mirroring an unstarted image would reintroduce that, one registry over.

It then **verifies against the registry** rather than trusting the push: the digest ECR serves must equal GHCR's, and the manifest must still list both architectures. It also fails if the tag list was empty — a mirror that copied nothing would otherwise report success.

## Inert until configured

With `AWS_ECR_ROLE_ARN` and `ECR_PUBLIC_REGISTRY` unset the `if` is false and the job is skipped, so **merging this changes nothing** until the role exists. OIDC rather than an access key, for the reason `publish-crate.yml` already gives: no long-lived secret to rotate, leak or quietly expire.

To turn it on, two repository variables:

| variable | value |
|---|---|
| `ECR_PUBLIC_REGISTRY` | `public.ecr.aws/f9f6l5u4` |
| `AWS_ECR_ROLE_ARN` | the role ARN below |

and an IAM role trusting GitHub's OIDC provider for this repository, with `ecr-public:*` push permissions plus `sts:GetServiceBearerToken`. ECR Public's API is in `us-east-1` regardless of where images are served.

## Verified, and what is not

Checked locally against the real `ghcr.io/samyama-ai/samyama-graph:1.8.0`:

| | result |
|---|---|
| tag loop over a metadata-action `tags` block | yields the 4 expected tags (`1.8.0`, `1.8`, `1`, `latest`) |
| digest format string | `sha256:a0ae9d01…` |
| platform extraction | exactly `linux/amd64 linux/arm64` — attestation manifests correctly filtered |
| workflow YAML | parses; jobs `build-and-push`, `mirror-to-ecr`, `smoke-test` |

**One thing I could not verify without the role, and say so in the workflow rather than leaving it to be discovered during a release:** the GHCR index carries two platform manifests *and* two attestation manifests (`unknown/unknown`, `vnd.docker.reference.*`). Whether ECR Public accepts those is not testable from here. If the push is rejected on them, the comment in the workflow gives the fallback — copy the two platform manifests explicitly, which keeps attestations on GHCR where the README sends people.

## Worth deciding alongside this

Automating the mirror is one option; **retiring ECR is the other**, and this PR does not assume the answer. GHCR is already automated, already has 1.8.0, and the main README moved to it in #1252. Either way the 70 stale `1.1.0` references need a pass — the only question is which registry they end up pointing at. If ECR is retired, this can be closed unmerged.


Kalyanpulikanti-01:
Pushed `3d4fe8c` in response to the review concern about this being a public repository. Two changes, plus the reasoning recorded in the workflow itself.

**What changed**

- The job now runs under `environment: ecr-public`, matching `publish-crate`, `publish-npm` and `publish-pypi` — it was the only publish job without one. The role ARN is an environment secret, reachable only from a job that names that environment, and a required-reviewer rule can be added to the environment later without editing this file.
- The ARN moved from `vars` to `secrets`. Repository variables are printed verbatim in public run logs and the ARN carries the AWS account id. Nothing about an ARN authenticates on its own — it names a role, it does not open it — but there is no reason to publish the account id. The `if` gate stays on `vars.ECR_PUBLIC_REGISTRY`, because the `secrets` context is **not available in a job-level `if`**; gating on a secret would evaluate empty and skip the mirror on every release without saying why.

**Why an outsider cannot reach this**

| | |
|---|---|
| Fork PR | This workflow has no `pull_request` trigger at all — only `release: published` and `workflow_dispatch`, both of which require write access. Separately, GitHub gives a fork PR neither secrets nor an id-token. |
| A malicious PR editing this file | Same answer: the edited workflow does not run on the PR, and if it somehow did, the OIDC subject would be a PR ref, which the trust policy refuses. |
| A branch in this repo | The trust policy accepts `repo:samyama-ai/samyama-graph:ref:refs/tags/v*` only. A branch build presents a different subject and AWS declines it. |
| Log exposure | `get-login-password` is piped directly into `docker login --password-stdin` — never a variable, a file or an echo. `configure-aws-credentials` registers what it issues as masked. The script uses `set -euo pipefail`, not `set -x`. |
| If a run were compromised anyway | The role can push to `repository/samyama-graph` and nothing else: no delete, no repository creation, no private ECR, no other service, and the credential expires with the run. The blast radius is a bad image tag — which is equally true of the GHCR push in the job above. |

**Residual risks, stated rather than implied**

1. Anyone with write access can dispatch this at a tag and publish an image. That is inherent to CI publishing and is the same for crates.io, npm, PyPI and GHCR today.
2. `PutImage` can overwrite an existing tag, including `latest`. Delete was deliberately left out of the policy, so an image cannot be removed, only superseded.
3. Adding a required reviewer to the `ecr-public` environment would make every ECR push need a human approval. None of the existing publish environments have one today, so I have not added it unilaterally — it is a one-click change in environment settings if you want publishes to be approved.

PR #1416, opened 2026-09-21
tag-release: say why the release-gate fetch failed, instead of guessing

Kalyanpulikanti-01:
The gate step discarded git's stderr and then asserted a single explanation for every failure:

```bash
if ! git fetch origin release-gate --depth=1 2>/dev/null; then
  echo "::error::no release-gate branch — the nightly has never published a verdict"
```

**Today that message blocked v1.9.0 while the branch existed and carried a green, in-date verdict.**

```
$ git ls-remote origin refs/heads/release-gate
88ae8e2b5b96469d8929c0438b63ba24a6a8f1a1  refs/heads/release-gate
```

The verdict on it: `"status": "pass"`, measured 2026-09-17, well inside the 8-day limit.

## What was ruled out

The same command succeeds from a workstation against the same remote under every configuration I could match to the runner:

| configuration | result |
|---|---|
| full clone, wildcard refspec | fetches |
| blobless clone, `--branch main` | fetches |
| single-branch refspec, as `actions/checkout` configures | fetches |
| with and without `--depth=1` | fetches |
| the step's exact shell (`set -uo pipefail`, `if !`, `2>/dev/null`) | succeeds |

Two workflow runs failed identically (35607105345 and its re-run), so it is not transient. The cause is specific to the runner — and it was thrown away before anyone could read it.

## What this changes

Nothing about the fetch. It stops the step lying about why it failed:

- a genuinely missing branch reports exactly what it reported before;
- anything else reports git's own error text.

Checked against the real remote on both paths: the success path, and a nonexistent branch correctly reporting "no release-gate branch".

## Why it matters beyond today

This is the step that gates every release. An error that names a wrong cause sends the next person to the benchmarks repo to investigate a nightly that is working fine — which is where today went before `ls-remote` settled it. The next failed run will name the real reason in its own log.

Worth noting the shape: `2>/dev/null` plus a hardcoded explanation is the same failure this repo has hit before — a check that states a conclusion its evidence does not support. `check_release_gate.py`'s own docstring makes the point about `partial` versus `pass`.


Kalyanpulikanti-01:
Closing this — the premise is wrong, and the PR should not stand on a public repo asserting it.

**The fetch never failed.** Re-running the job with debug logging shows the verdict being read correctly:

```
{ "suite": "CH-REGRESS", "status": "pass", "measured_at": "2026-09-17T20:21:25Z",
  "engine_commit": "92cf07d", ... }
##[error]CH-REGRESS passed, but not on this build: 51 commit(s) changed engine code
```

`actions/checkout` with `fetch-depth: 0` had already fetched `release-gate` before the step ran, and the step's own fetch then succeeded too.

**What misled me:** `gh run view --log-failed` prints the step's script source as well as its output, so the line `echo "::error::no release-gate branch — the nightly has never published a verdict"` appeared in the log as *source text*, and I read it as an emitted error. The real error was two lines further down. The step's error handling is fine as written.

The actual blocker is what it always was, and what the workflow said plainly: the verdict measured `92cf07d`, and the commits merged since have not been measured. That needs a CH-REGRESS run against the commit being tagged — not a code change here.

